PR #12, opened 2017-05-02
Add 4 Pack Sealed and Modern-Only Chaos

tritoch:
Adds checkbox to:
- Sealed for 4-pack sealed
- Chaos for modern-only

### DIFF
--- a/src/pool.js
+++ b/src/pool.js
@@ -24,7 +24,7 @@ function pickFoil(set) {
   var rngFoil = _.rand(6)
   if (rngFoil < 1)
     if (set.mythic)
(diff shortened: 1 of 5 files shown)


src/pool.js
```diff
@@ -100,6 +100,10 @@ function toPack(code) {
       special = selectRarity(set)
       foilCard = true
       break
+    case 'MM3':
```
ZeldaZach: You can probably combine all of these cases

case 'EMA':
case 'MMA':
etc

tritoch: Neat

PR #29, opened 2017-07-30
close #19 - Show progress in draft - pack and pick number

### LINKED ISSUE #19: Show progress in draft - pack and pick number

HerveH44:
Actually this is just to show the pack number.
I just have a question : the value of App.state.round may be -1 sometimes, but it means the game is finished. Should I take it in consideration ?
Thanks in advance,
Herve

### DIFF
--- a/public/src/components/grid.js
+++ b/public/src/components/grid.js
@@ -32,7 +32,7 @@ function zone(zoneName) {
 
   return d.div({ className: 'zone' },
     d.h1({}, Spaced(
-      d.span({}, `${zoneName} ${zoneName === 'pack' ? App.state.round : ''}`),
+      d.span({}, `${zoneName}${zoneName === 'pack' ? " " + App.state.round : ""}`),
       d.span({}, `${cards.length} ${zoneName === 'pack' ? ' /  ' + cards[0].packSize.toString() : ""} ${cards.length === 1 ? 'card' : 'cards' }`))),
     items)
 }
--- a/public/src/components/grid.js
+++ b/public/src/components/grid.js
@@ -33,6 +33,6 @@ function zone(zoneName) {
   return d.div({ className: 'zone' },
     d.h1({}, Spaced(
       d.span({}, `${zoneName}${zoneName === 'pack' ? " " + App.state.round : ""}`),
-      d.span({}, `${cards.length} ${zoneName === 'pack' ? ' /  ' + cards[0].packSize.toString() : ""} ${cards.length === 1 ? 'card' : 'cards' }`))),
+      d.span({}, `${zoneName === 'pack' ? 'Pick ' : ''}${cards.length} ${zoneName === 'pack' ? ' /  ' + cards[0].packSize.toString() : ""} ${cards.length === 1 ? 'card' : 'cards' }`))),
     items)
 }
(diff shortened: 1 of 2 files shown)


tooomm:
It looks like that if you try it on the beta right now:
packcards
packcards2
packcards3
Could you use the word `pick` instead `cards`? I like that you display the maximum card number as well, that might be handy for cubes or custom packs.
For example: `PACK 2 • PICK 8/14` 

<br>

>I just have a question : the value of App.state.round may be -1 sometimes, but it means the game is finished. Should I take it in consideration ?

@dev-id can you help him here?

HerveH44:
> Could you use the word pick instead cards? 
 
Sure, I can do that. Can you show that as a problem for the PR ? I don't know how it works, but I saw on git-hub that you can mark things as problematic for the merge of a PR. 

FYI, I'm actually without Internet at home for 2 weeks, so don't expect a fix before the ~20 of September!

public/src/components/grid.js
```diff
@@ -32,7 +32,7 @@ function zone(zoneName) {
 
   return d.div({ className: 'zone' },
     d.h1({}, Spaced(
-      d.span({}, zoneName),
+      d.span({}, `${zoneName}${zoneName === 'pack' ? " " + App.state.round : ""}`),
       d.span({}, `${cards.length} ${zoneName === 'pack' ? ' /  ' + cards[0].packSize.toString() : ""} ${cards.length === 1 ? 'card' : 'cards' }`))),
```
tooomm: >Could you use the word pick instead cards? Sticking with the established jargon is helpful imo.
>I like that you display the maximum card number as well, that might be handy for cubes or custom packs.
>For example: `PACK 2 • PICK 8/14`

tooomm:
With the recent change it looks like this:
pp
(checked on https://beta-dr4ft.herokuapp.com/, thanks for hosting @HerveH44!)

Did you stick with the "cards" term at the end for a special reason?
I think I would prefer a simple `Pack 1 • Pick 12/14` instead. We don't have to overload the interface, this information should be obvious and easy to figure out, no?

HerveH44:
Hey Tom,

You're right; I didn't check the "cards" at the end ... sorry!
I was thinking; isn't just better to say : Pack X - Pick Y ? (And not like you said Pack X - Pick Y/Z) Normally, we always use this denomination ("first pick first pack, i took that ...). 

I'll update that on the beta, tell me what you think.
Ciao,
Hervé

tooomm:
That looks fine and was my initial request 👍 
p1p4

But later I realized
>that (displaying) the maximum card number (...) might be handy for cubes or custom packs.

Because otherwise you don't know how much cards are in the pack. Normally it should be ~15, but in cubes with custom packs (or other card games!) it might be completely different.

So I see two options here:
- we keep the max cards in a pack behind pick number
--> `Pack 1 • Pick 4/14`
- we go with the simple and clean pack x - pick y, and then open an issue about missing max card in pack information (could be displayed with the game mode somehwere)
--> `Pack 1 • Pick 4`

What's your call @dev-id? I think I tend to the second and we find a better place for that, it's not the most important information and doesn't need that prominent placement.

Issue #23, opened 2017-05-20
host: possibility to change seating


tooomm:
A host should be able to swap the position of players before the draft starts (with little arrows for example).
This would allow for easier team drafts which have a fixed seating.
Also, when an external site/service requires a specific player order for example or there is any other reason you want to influence things (redo a draft with same order...) it would be nice to be more flexible.

tooomm:
https://github.com/dr4fters/dr4ft/pull/33 adds initial support for this 👍 

Some issues still stand with that implementation:
1) >What I miss there is, that once a host start swapping people around with the arrow buttons, shuffle should get automatically checked and disabled to make sure his preferred seating order isn't mixed up again!
2) >Regarding the swapping buttons itself:
Since you can't cycle a player on position 1 to the bottom if you hit <kbd>up</kbd>, It would be clearer to disable/grey out <kbd>up</kbd> for the first player, and <kbd>down</kbd> for the last.
You can hit those buttons but just nothing happens currently.

I think we also talked in gitter chat about the last one and said that cycling players from 1 "up" to 8 would also be very nice. So it's either of those: Allow cycling or disable up/down buttons on first/last position. I prefer cycling since it's more intuitive.

PR #33, opened 2017-08-12
Dev: Add possibility to change seating (#23) and possibility to shuffle

HerveH44:
This pull requests makes it possible to change the seat of players, but using 2 buttons (up and down), that i inserted just before the button "kick".

Doing so, I also needed to add a button "Shuffle players", to control that the seats remain the same even after game started. 

Thank you for your reviews.


tooomm:
`Shuffle` means here randomizing the order of people once the draft starts (at the beginning)?
That used to be always like that, right? We just have an option now for that?

HerveH44:
> The new shuffle player tick box controls if people get randomizing at the beginning once the draft starts?
That used to be always like that, right? We just have an option now to disable that with your change?
>
That's exact. The Change Seating button (I don't know if you noticed it as a host of a draft) forces me to disable the "auto-shuffle" behavior. 

I added that particular "shuffle players" button which directly actives or deactivates the shuffle of the players at the start of the game. Because, if someone changes seats, I guess it doesn't want them shuffle when the game starts.
What do you think ? Is it clear enough button ?

tooomm:
>What do you think ? Is it clear enough button ?

It's ok. Of course I realized the arrows to swap people around, too. But I wanted to make sure about shuffle first. What I miss there is, that once a host start swapping people around with the arrow buttons, `shuffle` should get automatically checked and disabled to make sure his preferred seating order isn't mixed up again!

Regarding the swapping buttons itself:
Since you can't cycle a player on position 1 to the bottom if you hit `up`, It would be clearer to disable/grey out `up` for the first player, and `down` for the last. You can hit those buttons but just nothing happens.

Issue #34, opened 2017-08-16
Draft Log


HerveH44:
Hi, 

I've been trying to make a draft log from the app. I'll suggest a pull request, tell me if it's fine ! 

I tried, and it works with the site : http://magic.flooey.org 
I contacted them, and they replied that my file is fine. 

Let me know, what you think about it,

Herve

tooomm:
Do you have this change running on a site so that i can see it in action?

HerveH44:
Actually no, I'm trying to put it on Heroku but it's not that simple. I'll try in the next few days but I'm quite busy at the moment. 

It's easier to create on your own computer, I don't know if you're familiar with it.

If you want, I could add some pictures and log files to give a bit of a hint.

tooomm:
Just tested it with a draft on the beta site. It works nicely and is a charm to use on that page.
Too bad it's just your data and not from all drafters.

This issue can be closed once you update your main site @dev-id ?

tooomm:
The main site is updated in between. This can be closed.

PR #35, opened 2017-08-16
Dev: add draft log

HerveH44:
This is the answer of #34 

As you can see, there's still some to do (maybe control the name of the file, visibility - as it's seen also during the game and when we don't draft ). Any help would be appreciated :) !

Thanks,


ZeldaZach:
Good start, I'll get this in for now

PR #56, opened 2017-09-21
Add Docker

HerveH44:
Hi, 
This my commit to a dockerfile!

I start from node-alpine because it's a smaller build.

There's a bit of a problem with hardtyping the listening port though. I don't how to set it as a variable (got from building env for example). 

Bye,

Hervé


HerveH44:
PR Useless since the other PR #84 has been merged

PR #57, opened 2017-09-21
Add Ixalan set

HerveH44:


dev-id:
Looks good. Nice clean reference for anyone adding sets in the future.

tooomm:
There could be a wiki page about basics like that, too. :)

PR #60, opened 2017-09-28
restructure readme

tooomm:
re-organized and re-grouped phrases

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,21 +2,25 @@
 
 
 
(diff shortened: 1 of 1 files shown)


HerveH44:
It still lacks an important thing, in my opinion : 
In order to install the app (before make run), one needs to launch make 'install'

By itself, make run will not launch the app :(

tooomm:
Just changed the place where sentences show up basically.

Can you add that in a new PR, please? @HerveH44

PR #62, opened 2017-09-29
Close #52 - Managing Doublefaced cards

### LINKED ISSUE #52: add related card pictures (dfc)
I highly like to see the back sites of Double-Faced-Cards in the preview window!

Otherwise you have no way to know about the back site in draft or during building of limited pools.
So this information is very critical, especially for new sets or cubes with older or less common cards.

It could look like this:
- normal DFC (transform, flip)
dr4ft_dfc
- special DFC (melt)
melt
Where the both halves actually show the combined melted card in normal card size.

Back sides should be displayed a bit transparent for all of them.

<br>

`allsets.json` from mtgjson holds useful information about the relevant cards.
In the "layout" key they are tagged with "meld" or "double faced" and have the names of the other sites/cards, in that object too.

HerveH44:
This pull adds 
1. show flipped face during mouseOver on doublefaced cards
2. show the two faces of double-faced cards on main and and side zones, thanks to imageHelper popup.

### DIFF
--- a/public/src/components/cols.js
+++ b/public/src/components/cols.js
@@ -5,21 +5,19 @@ let d = React.DOM
 export default React.createClass({
   getInitialState() {
     return {
-      className: 'right'
+      className: 'right',
     }
   },
   render() {
     let zones = this.props.zones.map(this.zone)
-    let img = this.state.url && d.img({
-      className: this.state.className,
-      id: 'img',
-      onMouseEnter: this.enter.bind(this, this.state.url),
-      src: this.state.url
+    const img = this.state.card && React.createElement(ImageHelper, {
+      card,   
+      className 
     })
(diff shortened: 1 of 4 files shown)


tooomm:
Gave this another check on your beta on heroku - looks damn sweet. And works nicely 👍

HerveH44:
We can add it if you want! 

FY, I stil didn't add any PR in the refactoring ... I guess It shouldn't take too long to rewrite the same code using JSX ( I even guess that it should take very little time ).

tooomm:
Ok, so this pr is up to date and ready. 👍 

Maybe it's best to wait for the refactor to finish. It looks like you're actively working on it.
We don't have to include code that has to change again soon.

PR #73, opened 2018-01-02
Additional sets to ignore

dev-id:


ZeldaZach:
This should be a const in a header location

PR #83, opened 2018-01-17
Update game.js

sammorrow:


ZeldaZach:
Thanks for attempting to fix something, even though it seems fine. Why don't you check out our Issues backlog for some things that may need attention? :)

sammorrow:
Sorry, accidentally pushed my changes up here. Couple of friends wanted to have the draft room last for like a day instead of an hour. I figured you wouldn't want that option on a public server so I'm working on a fork instead.

PR #130, opened 2018-08-08
Dev: add M19 and basic cards; closes #122, closes #119 

### LINKED ISSUE #119: M19 dual land in packs
Hey,. what heuristic are you using for M19 "common" duals?   It should be in pool.js but it's not been pushed for a while

### LINKED ISSUE #122: m19 has wrong name
Core Set 2019 > Magic 2019

HerveH44:
Add the support for M19

### DIFF
--- a/public/src/sets.js
+++ b/public/src/sets.js
@@ -79,6 +79,7 @@ export default {
     "Arabian Nights": "ARN"
   },
   core: {
+    "Magic 2019": "M19",
     "Magic Origins": "ORI",
     "Magic 2015": "M15",
     "Magic 2014": "M14",
--- a/public/src/sets.js
+++ b/public/src/sets.js
@@ -79,7 +79,7 @@ export default {
     "Arabian Nights": "ARN"
   },
   core: {
-    "Magic 2019": "M19",
+    "Core Set 2019": "M19",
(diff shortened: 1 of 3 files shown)


public/src/sets.js
```diff
@@ -79,6 +79,7 @@ export default {
     "Arabian Nights": "ARN"
   },
   core: {
+    "Magic 2019": "M19",
```
tooomm: Regarding #122 it should be named `Core Set 2019`, see https://magic.wizards.com/en/products/core-2019

src/pool.js
```diff
@@ -132,6 +132,12 @@ function toPack(code) {
       else
         special = special.common
       break
+    case 'M19':
+      //http://wizardsmagic.tumblr.com/post/175584204911/core-set-2019-packs-basic-lands-and-upcoming
+      const dualLand = _.rand(12) < 6;
+      pack.push(_.choose(1, dualLand ? special : basic))
+      special = false; // to desactivate the pick of special card
```
tooomm: deactivate?

HerveH44:
Ok, should be good now. I simplified to keep good rarity in frontend. Now I just filter during the creation of the booster all the dualLands from commons and add one on basis of 5/12.

PR #138, opened 2018-08-09
Dev: add BBD support; Closes #120

### LINKED ISSUE #120: Missing Battlebond BBD for use with cube
Unable to start a room with a cube list with two BBD cards.

HerveH44:
Add BBD cards. 

Though we don't should the set as draftable (it should be only for Two-Headed Giant)

### DIFF
--- a/src/make/cards.js
+++ b/src/make/cards.js
@@ -19,7 +19,6 @@ before();
 
 var types = ["core", "expansion", "commander", "planechase", "starter", "un"];
 var codes = ["EMA", "MMA", "VMA", "CNS", "TPR", "MM2", "EXP", "MPS", "CN2", "MM3", "MPS_AKH", "IMA", "BBD"];
-
 for (var code in raw) {
   var set = raw[code];
   if (types.indexOf(set.type) > -1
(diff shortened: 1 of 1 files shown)


ZeldaZach:
If people want to draft it for chaos or smth, let em :)

tooomm:
I was thinking the same. It doesn't really hurt to list them I guess.
Which actually applies to commander sets or all sets in general...

src/make/cards.js
```diff
@@ -18,7 +18,7 @@ var setsToIgnore = ['TSB','ITP','CP1','CP2','CP3'];
 before()
 
 var types = ['core', 'expansion', 'commander', 'planechase', 'starter', 'un']
-var codes = ['EMA', 'MMA', 'VMA', 'CNS', 'TPR', 'MM2', 'EXP', 'MPS', 'CN2', 'MM3', 'MPS_AKH', "IMA"]
+var codes = ['EMA', 'MMA', 'VMA', 'CNS', 'TPR', 'MM2', 'EXP', 'MPS', 'CN2', 'MM3', 'MPS_AKH', "IMA", "BBD"]
```
tooomm: Wondering why masters sets are not filtered by `masters` type, but some are listed by their codes instead?!

tooomm:
What are you thoughts @HerveH44 ?

- https://github.com/dr4fters/dr4ft/pull/138#issuecomment-411874992

- https://github.com/dr4fters/dr4ft/pull/138#discussion_r209069676

tooomm:
Superseded by #169

PR #145, opened 2018-08-10
Dev: add webpack-cli

HerveH44:
Add a new dependency which became mandatory following the bump of webpack

PR #150, opened 2018-08-10
CI: enable travis

tooomm:
First version of working Travis CI:
- Install test on Node.js, which should had spotted the need for https://github.com/dr4fters/dr4ft/pull/145.
- Added ESLint check
- Added Node.js LTS
- Added check for Dockerfile

New GitHub Checks API in action: https://github.com/dr4fters/dr4ft/pull/150/checks

Note, that there is currently still a bug with the new Travis App for GitHub and builds are reported and shown twice, tracked in https://github.com/travis-ci/travis-ci/issues/9618:
>Yes, right now we're reporting the status two times per build triggered, one for the previous commit statuses and another one via GitHub Checks.
>
>We are aware of this issue, and we are working to clean it up in the coming weeks.
>
>Thank you for your understanding and patience.

---

This PR should be merged before #146

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
       node_js: node
       cache:
         directories:
-        - node_modules
+          - node_modules
       install:
         - npm install
       script: skip    # no tests configured yet
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ env:    # needed as empty for allow_failure on env to work, see https://docs.tra
 
 matrix:
   include:
-    - name: Install Node.js app on newest Node.js
+    - name: Install Node.js app (newest Node.js)
       language: node_js
       node_js: node
       cache:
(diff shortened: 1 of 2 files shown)


tooomm:
@HerveH44 can you please configure Heroku Review App so that it only starts deploying after the checks passed? I think that makes most sense now.

`“Wait for CI to pass before deploy”`

tooomm:
@HerveH44 I also let travis run eslint now.
Can you double check the eslint configuration you introduced?
https://github.com/dr4fters/dr4ft/blob/master/.eslintrc.js

It looks like our codebase doesn't match it and travis reports 1300 warnings/errors when running `eslint --ignore-path .gitignore .`...
https://travis-ci.com/dr4fters/dr4ft/jobs/139450726#L519

tooomm:
- Added test against LTS node version
- Moved ESLint checks to a seperate build which is allowed to fail (Travis will only consider the other two build when deciding to show failing or pass), so eslint is no road blocker for pr's and we can fix as we see need or have time.

Example: https://travis-ci.com/dr4fters/dr4ft/builds/81588804

PR #152, opened 2018-08-11
eslint-ing

tooomm:
@HerveH44 please double check this

I only run `eslint --fix`, `npm audit --fix` and updated the *.eslintrc.js* file to removed a hint about deprecated options.

As can be seen [here](https://travis-ci.com/dr4fters/dr4ft/jobs/139509374#L454), this fixes most of the errors/warnings already.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,7 @@ module.exports = {
       "plugin:react/recommended"
     ],
     "parserOptions": {
-        "ecmaVersion": 2018
+        "ecmaVersion": 2018,
         "ecmaFeatures": {
             "jsx": true
         },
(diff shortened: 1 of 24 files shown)


tooomm:
We also use two deprecated methods in https://github.com/dr4fters/dr4ft/blob/master/public/src/game/Game.jsx

>  17:16  error  componentWillMount is deprecated since React 16.3.0, use UNSAFE_componentWillMount instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillmount
react/no-deprecated
  17:16  error  componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops
react/no-deprecated

tooomm:
See #159

PR #155, opened 2018-08-12
draftStats

benjaminpwagner:
I have added some functionality that will allow the server to keep track of users pick orders in draft. I would like to use this data to create some kind of format analyzer that is able to rank a cards pick priority in a given format. The options are endless, this data could also enable some kind of machine-learning draft bot.

Probably wouldn't be ideal to leave the json files in data/draftStats, but instead store them in a database or upload them to the server I will be hosting the analysis tools on.


HerveH44:
Just a quick question: have you looked at the info in draftLog? A long time ago, I did a feature to enable the log of a draft with the card taken among the cards in the pack. Could it be possible for you to plug in these data instead of maybe duplicating it?

benjaminpwagner:
@HerveH44 Yes if you look at my line #140 of human.js, I am using the data from draftLog as an argument for updateDraftStats.

benjaminpwagner:
@tooomm ./data/cap.json only stores what the players have picked. I am interested in their decision making throughout the draft process - what cards they picked over other cards and what cards they already had in their pool when they made the decision.

This will help in figuring out a formats pick order and a hopefully a bot that can draft while making smart decisions.

ZeldaZach:
Do we want to do anything with this? looks like a nice thing.
@HerveH44 @dev-id

PR #159, opened 2018-08-14
run eslint + npm audit

tooomm:
This supersedes https://github.com/dr4fters/dr4ft/pull/152.
It's the same, just with cleaner commits:

>I only run `eslint --fix`, `npm audit fix` and updated the *.eslintrc.js* file to removed a hint about deprecated options.


tooomm:
Simple change, merging to have a cleaner and secure base.

PR #178, opened 2018-08-23
Fix: adjust keys of JSX arrays to stop the app from refreshing too much

### LINKED ISSUE #173: react refreshes components too often
The components are refreshed based on their key. At the moment the keys of mosts components are created using "uniqueId" of lodash. This creates unnecessary refresh of the components that tend to create freezes or worse. 

Solution: get rid of uniqueId as much as possible and use the index of arrays (as most of the times the components are created for arrays of data) to keep the "key" variable of the components

HerveH44:
Closes #173

### DIFF
--- a/public/src/game/Cols.jsx
+++ b/public/src/game/Cols.jsx
@@ -57,8 +57,8 @@ const Zones = ({onMouseEnter, zoneNames, onMouseLeave}) => {
     let cols = [];
 
     for (let key in zone) {
-      let items = zone[key].map(card =>
-        <div key={_.uid()}
+      let items = zone[key].map((card, index) =>
+        <div key={index}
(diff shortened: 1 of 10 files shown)


tooomm:
Nothing I can do here. I also pinged Zach in Gitter a few days ago...
If you feel comfortable that everything is fine, go ahead and merge @HerveH44.

HerveH44:
@ZeldaZach As @tooomm indicated, I'll merge this PR this week-end if you don't have any problem with that. 

In the contrary, tell me what I can do better :) ! 

Honestly, I think it's quite a good improvement. I mean that for now, even the "image Helper" doesn't stick more than one second when you hover on your main's columns. It's pretty annoying (imo).

PR #179, opened 2018-08-23
Dev: adjust draftlog filename

### LINKED ISSUE #172: "Download deck" and "Download draft log" should share the same naming scheme
To easily match them it would be very helpful if both use the same naming pattern.

Right now only the deck name can get changed before downloading, the draft log is always fixed to a pattern with UST? time appended.

HerveH44:
Closes #172 

Now the filename is took from the App filename. We just append -draftlog.txt at the end

### DIFF
--- a/public/src/cards.js
+++ b/public/src/cards.js
@@ -118,10 +118,9 @@ let events = {
     App.state.log = draftLog;
   },
   getLog() {
-    let {id, packsInfo, log, players, self, sets, type} = App.state;
+    let {id, log, players, self, sets, type, filename} = App.state;
     let isCube = /cube/.test(type);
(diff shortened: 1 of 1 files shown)


public/src/cards.js
```diff
@@ -140,7 +139,7 @@ let events = {
       });
     }
 
-    _.download(data.join("\n"), filename);
+    _.download(data.join("\n"), `${filename}-draftlog.txt`);
```
tooomm: Did you change the file extension to `.txt` for a particular reason?

`.txt` and `.log` both work on https://magicprotools.com/draft/upload

tooomm: @HerveH44

HerveH44: Ehh, i just imagined that it woul be clear that it's a txt. No particular reason no!

Issue #185, opened 2018-08-30
dr4ft.com and dr4ft.info not working


Rendroc:
Everytime I've tried to start a cube Sealed game, it gets stuck and I see no cards. My opponent apparently see their card pool but not me, and I'm the creator.

HerveH44:
Hi, 

If you've played before on these sites. It may be linked with your browser's cache. Try to delete the cache of your browser and then retry. It may also be possible to do a "hard-refresh" directly on the webpage, when you are visiting one of these sites, using the keys "ctrl + F5" or even "ctrl + shift + F5".

Tell us if that worked for you, 

Hervé

Rendroc:
Nothing is working! I tried different browers, I cleared Cache and cookies, and tried the hard refresh command, and nothing works. This is so strange.

tooomm:
What's your exact environment and what did you exactly do? When does it get stuck?

gsainsbury86:
I've been getting similar behaviour since switching ISPs a couple of weeks ago.

I see this error in the console. (Win 10 - Chrome)

`engine.io.js:4387 WebSocket connection to 'ws://dr4ft.com/engine.io/?id=6zeh4jwjehh&name=dr4fter&EIO=3&transport=websocket&sid=sCOIzjf6QvdqC0ZIAAFv' failed: Error during WebSocket handshake: Unexpected response code: 400`

HerveH44:
Ok! Thanks for the console log. It should be because your ISP blocks Websocket traffic. It's possible, it happens sometimes that the router/gateway is configured in such a way that it blocks WS traffic. 

It's weird because I thought that engine.io was able to chage the transport if websocket was failing. I'll have a look later about this.

gsainsbury86:
Great, thanks. I have to use a silly router because my ISP has mac address filtering on it. I'm switching ISPs soon, so it won't affect me anymore hopefully but at least this sheds some light on the issue hopefully.

JineteDV:
![screen shot 2018-08-31 at 2 45 31 pm](https://user-images.githubusercontent.com/6514656/44930526-984ab800-ad2c-11e8-9894-d10bf0823110.png)
This should be a draft (not sealed) with one other person. macOS 10.13.6, Chrome, Java v8 update 181.

Rendroc:
[...]

PR #215, opened 2018-10-02
Bypass browser cache

HerveH44:
As #185 showed, people have problems when we update the app, because their browser caches the javascript and can creates problem. 

I updated the webpack configuration to create everytime a new app bundle with an hash based on the app file content. Doing so, the browser will think it has to download a new file every time we update our websites and it will no longer keep old cache.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ data
 public/out
 public/lib
 public/src/config.js
+public/index.html
 
 config.client.js
 config.server.js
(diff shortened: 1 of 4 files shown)


HerveH44:
I'll still be working on this feature for couple days/weeks. I think it's not enough, we have to do that also for react and react-dom ... 

In fact, we should also use a "build" directory instead of the public one. Right now, anything in the public directory is available, which means you can see our source code in any client, going to the url .../src/... !!

PR #218, opened 2018-10-05
Move Webpack build dir to specific /built/ directory

HerveH44:
Second improvement of bypassing browser cache when updating the website. 

Now I move the entire "public" directory to "built/" during bundle creation by webapck. At the end of the installation (npm install), the "build" directory is created and the app now points to it to deliver the static content. 

I think that might help a lot of people who suffer from the browser cache bugs.


tooomm:
What do you think @ZeldaZach?

ZeldaZach:
Seems fine to me

PR #223, opened 2018-10-13
Fix grn guildgates

HerveH44:
Fix the guildgates names, removing the (a) and (b) from their names


tooomm:
Isn't that fixed at mtgjson site already? @ZeldaZach

ZeldaZach:
V4, yes. V3, nope.

tooomm:
Looking into using v4 instead makes more sense then I guess?
Is there a current output available somewhere @ZeldaZach?

HerveH44:
@ZeldaZach V4, as outputed elsewhere (don't remember if it was on gitter), still lacks "booster" parameter on sets, and still doesn't show "multiverseId" of cards. 
I can't use v4 unless this is fixed :( !

ZeldaZach:
`multiverseId` was added back, but booster is missing... We need to come up with a better solution / way to impl it. Feel free to help chime in

ZeldaZach:
It should be there... it’s on my build. Are you sure you’re up to date with master

HerveH44:
Well after a quick check, there's indeed a "multiverseId" in v4 ... but v3 wrote it "multiverserid"

HerveH44:
So, we need to do a special PR to update dr4ft to v4.

PR #315, opened 2019-02-15
Use filepond as custom set handler

HerveH44:
Attempt to use [filepond](https://pqina.nl/filepond/) to handle custom set upload.

It's nice but a little bit big (~50kB) to have a nice way to upload custom sets.


ZeldaZach:
Doesn't seem to work on heroku... doesn't update the dropdown menus

HerveH44:
Yeah I don't know why. There's no log file or anything in the container ... But it works on my workstation. Doesi it work on yours?

PR #320, opened 2019-02-18
Refresh card image after click 

### LINKED ISSUE #314: Refresh card image (mouse over) after click
Happens with Firefox and Chrome on dr4ft.info

When you click a card in the piled view to move it to e.g. sideboard, the pile gets updated and the card beneath your mouse changes but the card preview does not refresh.

HerveH44:
Closes #314

### DIFF
--- a/public/src/game/Cols.jsx
+++ b/public/src/game/Cols.jsx
@@ -37,7 +37,7 @@ class Cols extends Component {
   render() {
     return (
       <div>
-        <Zones onMouseEnter={this.onMouseEnter} zoneNames={this.props.zones} onMouseLeave={this.onMouseLeave} />
+        <Zones onMouseOver={this.onMouseEnter} zoneNames={this.props.zones} onMouseLeave={this.onMouseLeave} />
         <ImageHelper onMouseEnter={this.onMouseEnter}
           className={this.state.className}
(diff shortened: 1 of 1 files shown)


tooomm:
Hmm, I see no difference between
https://dr4ft.info and https://beta-dr4ft-pr-320.herokuapp.com/

The preview pictures used to refresh when you click at the top card on the pile (the one you see completely).
But it doesn't work if you click on the bottom one (the one which name is only visible).

With this PR that behavior seems not to change?

Which change do you expect from this PR exactly @HerveH44 ?
I can then retest again 👍

HerveH44:
Have you tried to hardrefresh your browser ? It seems to work for me

HerveH44:
That's really weird, I can't find a way to understand how to fix this. There's no event on the card where the mouse is after the previous card disappeared... 

I feel that's out of my reach ....

tooomm:
Hmm... Why does it work when you click the top card, but not any other card?!

HerveH44:
That's a mystery .... ;) !

ZeldaZach:
What do we want to do with this?

HerveH44:
As for me, i'd push as it is. I can't do better than that ...

PR #321, opened 2019-02-18
Show DownloadLog button only when the game is finished

### LINKED ISSUE #180: draft log: disable download button until draft is finished
Before finishing a draft, the log file will be empty anyway and only contain the game id and player names.
It doesn't matter if you haven't picked anything yet, pack 1 is done or you are only missing the very last pick from pack 3... once the draft isn't finished there won't be any card picks included in the log.

HerveH44:
Closed #180

### DIFF
--- a/public/src/game/DeckSettings.jsx
+++ b/public/src/game/DeckSettings.jsx
@@ -133,7 +133,7 @@ const Copy = () => (
 );
 
 const Log = () => (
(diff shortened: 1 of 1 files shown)


HerveH44:
I couldn't just disable it. I don't know why though... I think it's good enough just to hide the button before the draft ends

PR #336, opened 2019-02-20
Sort packs by rarity by default

### LINKED ISSUE #330: Sort packs by rarity by default
Most draft apps sort by the packs by rarity by default.

Real life Magic is sorted this way, MTGO sorts this way, and Arena sorts this way. Please consider sorting this way by default, with the Mythic Rare/Rare cards first, uncommon next, then commons. Special slots, like Gates in the Ravnica sets should be last.

HerveH44:
Closes #330

### DIFF
--- a/public/src/app.js
+++ b/public/src/app.js
@@ -46,7 +46,7 @@ let App = {
     filename: "filename",
     filetype: "txt",
     side: false,
-    sort: "color",
+    sort: "rarity",
     log: {},
 
     get didGameStart() {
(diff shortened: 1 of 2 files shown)


tooomm:
👍 

But for RNA the gate comes right after the uncommon cards, not at the end.
You ordered special to the end, so there might be another issue.

HerveH44:
The gates are commons I think. They are treated as so. You would like another sort by CMC ?

ZeldaZach:
It’s a pointless case. They’re fine being in the commons area. Doesn’t have to be first.

Issue #77, opened 2018-01-03
display special cards better


tooomm:
Since https://github.com/dr4fters/dr4ft/pull/62 DFC's (`transform`) now display both sides in the preview and switch to the back side on hovering during picking as requested in https://github.com/dr4fters/dr4ft/issues/52. 👍 

However there are other special cards which are not covered yet
- [ ] Implement `meld` cards too, and show their combined back side next to each front side in the preview
Gatherer has their own multiverseID with their own picture for the combined meld card, so it's even easier than displaying 2 rotated pictures on top of each other: http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=439323
- [ ] Show both orientations (normal and 180° rotated) of `flip` cards from Kamigawa next to each other in the preview.
The flipped side of such cards is just unreadable and one has no clue during a draft what such a card does unless you know by heart... basically the same problem as with *transform* cards before.
Should also show the rotated card + fade it on hover during picking just like with DFC's now!

Or others which could optionally be displayed in an improved way for better view- and readability
- [x] Turn `split` and `fuse` cards 90° in the preview window
- [ ] For `aftermath` cards, have two pics in the previews next to each other same as DFC's have now.
One normal to see/read the main part; and one 90° rotated to see/read the aftermath part better!
This might not be ideal though because it eats up a lot of space and there might be a better way like using only the bottom "aftermath" half and display that rotated+scaled for example. Then it's like displaying two cards next to each other.

tooomm:
I want to give this a bump due to 
> * Turn `split` and `fuse` cards 90° in the preview window

So many sets have them nowerdays and it would be soo much easier to read them!

PR #337, opened 2019-02-20
Rotate card by 90 degree if card layout is fuse or split

HerveH44:
Answsers #77 split and fuse request.

If a card has a layout of "split" (includes also the fuse cards), we rotate the card by 90 degrees.

image


tooomm:
❤️ ❤️ ❤️

PR #343, opened 2019-02-22
Allow user to set card's size (to decrease loading image time)

### LINKED ISSUE #329: Delay with images on new pack
What is happening is the images aren’t loaded, so when the next pack comes, I have to wait for the images to load before I know what the card is, which eats into my timer.

Options:

1. Preload images from the next pack (bad idea).
2. Wait for all the images to load before displaying the pack and starting the pack timer.
3. Preload all the images from the set.
 
#1 is a bad idea because it could be used to cheat. #3 might be bad because it may be downloading way more cards than needed. #2 may be the way to go here.

HerveH44:
Closes #329 

image

### DIFF
--- a/public/src/app.js
+++ b/public/src/app.js
@@ -48,7 +48,7 @@ let App = {
     side: false,
     sort: "rarity",
     log: {},
-    cardSize: "small",
+    cardSize: "normal",
 
     get didGameStart() {
       // both round === 0 and round is undefined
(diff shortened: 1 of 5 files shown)


tooomm:
Not so sure about `Set cards image`.
What about `Card image quality`?

HerveH44:
@tooomm I pushed your propositions. And yes, it should closes #329

PR #345, opened 2019-02-22
Notify with toast a message when player has chatbox closed

### LINKED ISSUE #53: notifications for hidden chat window
Copied over from old old repo: https://github.com/aeosynth/draft/issues/131

>I sometimes miss the chat because I forget to check at all, or just don't check on a regular basis during a draft or a complex sealed pool building session.
You need as much space as possible for a nice building experience, so I end up hiding the chat... but also missing many (some important) messages!
>
>I would like to see the `Chat` text flashing for example or some other visual notification of new messages. Especially when the chat window is hidden.

More modern and interactive Web Notifications are also a viable consideration.

HerveH44:
Fix #53

peek 2019-02-22 16-39

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5504,7 +5504,7 @@
         },
         "readable-stream": {
           "version": "1.0.34",
-          "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-1.0.34.tgz",
+          "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-1.0.34.tgz",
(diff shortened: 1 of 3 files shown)


ZeldaZach:
We can also use a browser or OS built in notification system. See Facebook for example. This could be a fallback if they don’t want system notifications.

HerveH44:
Well browsers' notifications are when you don't look at the browser at all right? It seems a bit overkill for me to do so. Also you would receive a notification even when you're looking at the draft? That would be even weirder!

ZeldaZach:
Fair enough, this is cool!

PR #346, opened 2019-02-22
Timer hud + npm run start-dev

### LINKED ISSUE #16: constantly show pick timer as little overlay somewhere
Something like the card preview for example.

You always need to scroll up to check the timer right now.

Happens several times now to me:
You mark a card for autopick while re-considering other cards. To skip to the next pack you still double click on the card selected for autopick. But since the timer went down and/or due to the server delay I end up immediately picking the next card at the same position from the following pack.
If you always see the time, that won't happen.

benjaminpwagner:
Fix #16 

Locks pack timer to top left of screen when user scrolls down.

This way users can keep track of how much time they have left to pick when they are scrolled down the page.

Added nodemon (as a dev dependency) and npm run start-dev so server / client detects changes and reloads for dev purposes.


HerveH44:
Hey, what about webpack dev-server? Isn't it exactly the same thing? Look at "npm run dev-server"

ZeldaZach:
So i tested this out and I do like it, but like Herve said the CSS needs to be better. Maybe make it more noticeable / fancy so people know what they're looking at as I had to actively look for it.

benjaminpwagner:
@HerveH44 About the nodemon: As far as I know, webpack-dev-server only hot reloads for client side code change and has an identical effect to running webpack with a -w flag. Nodemon will restart the server in the event of server side code changes. 

Can someone attach a screenshot of what they see for the HUD timer? I'm on MacOS and chrome, so it make look different but on my screen it is quite evident where it is and what the timer indicates.

tooomm:
This is how it looks for me on Windows with Firefox:
win-ff


It's a nice improvement in my opinion, even though it's not perfectly stressed.

Something like the floating box in https://github.com/dr4fters/dr4ft/pull/345 in one of the upper or bottom corners would be more visible.

HerveH44:
I've never used nodemon but I'm more than happy to use it! Webpack dev-server is not useful anymore I guess!

benjaminpwagner:
So I had to use relative distances to get the timer hud where I wanted on scroll down.

Is it showing up here for everyone else?

image

tooomm:
There are several issues for me:
- timer in the middle of the page
timer

- timer freezes sometimes
timer2

- chat overlays and cuts timer off
timer3

benjaminpwagner:
Took a different approach, should be good cross browser/device. Let me know what you guys think.

ZeldaZach:
Interesting approach. Getting better, but not perfect. when you scroll down, the cards shift.

screenshot 2019-02-25 15 54 58
screenshot 2019-02-25 15 54 56

ZeldaZach:
Also...
screenshot 2019-02-25 15 56 15

benjaminpwagner:
Took a different approach again.

@ZeldaZach New commits should fix those issues + look slightly better.

ZeldaZach:
Getting better for sure! Still some issues, see screenshots

screenshot 2019-02-25 17 35 22
screenshot 2019-02-25 17 35 13
screenshot 2019-02-25 17 35 06
screenshot 2019-02-25 17 34 59

HerveH44:
image

benjaminpwagner:
Hm, looks like your browser is overriding the global font we are using on the site.

I'll try to adjust the width to be font-agnostic, but I could always find a fallback font that works with the size of the panel if that doesn't work.

tooomm:
This is turning out nicely!
I also tested it on a tablet with smaller screen, but higher resolution without issues.

What does change now is that the segments/blocks (Players, Game and Settings...) get regrouped once the chat is displayed on smaller screens. The boxes overlay each other a bit in that case:
timer

benjaminpwagner:
@tooomm The chat was cutting off the config boxes on smaller screens so we added a wrap around. I can give those elements a bit of margin to prevent any overlap.

PR #358, opened 2019-02-26
Handle double faced cards and flipped card

### LINKED ISSUE #77: display special cards better

### LINKED ISSUE #107: problem with dfc's
The back sides are not loading.
Experienced this with "Azor's Gateway" and "Journey to Eternity" on draft.info, both from RIX. But also "Legion's Landing" from XLN.

It looks like the back side of the card is treated as a card on their own, I randomly opened the back side of Journey, too:
front
back

HerveH44:
Closes #107 

I rechechecked how MTGJson v4 indicates special cards and modified how to catch double faced cards.

Closes #77 

I didn't handle aftermath cards because I'm really not sure how to handle them (the rotate 90° and crop half the card is not convincing enough IMO)


ZeldaZach:
So when I was testing this on the server i thought it was working, but when i scrolled over a flipped card it crashed the program. I built it locally but don't have the flipping mech on scroll over, so not sure what's up.

ZeldaZach:
Update: Works as expected. Heroku is being taxed too much

tooomm:
This is awesome, thank you @HerveH44 !!

PR #370, opened 2019-02-27
rephrase nickname msg

tooomm:
Wanted to clarify the location.

Opinions please.


tooomm:
@HerveH44 Why is there no heroku deployment made?

PR #375, opened 2019-02-27
update known pages

tooomm:
It looks like .com is down currently.

Any news if @dev-id is doing an update etc. or if the page is gone?

---

@ZeldaZach could you update the dr4ft page link in the repo to be https please?


ZeldaZach:
What do you need from me?

tooomm:
url

PR #384, opened 2019-03-06
Preset cubes dropdown

### LINKED ISSUE #3: Create preset cubes
Create a list of preset cubes such as the ones from MTGO and allow them to be easily loaded into the cube list.

benjaminpwagner:
Adds a dropdown next to the cube list textarea. Fills textarea upon selection (Using global App.state)

Also added axios and babel-polyfill - updated webpack config to allow for async/await usage on the client side.

Fixes #3 

update: fixes clean-webpack-plugin error with v2


HerveH44:
Hey, that seems awesome. Though I have couple questions: 

- why axios and not use builtin "fetch" ?
- why remove webpack clean plugin ? You don't think it's safer to delete all previous builds before making a new one?
-The preset cubes selects/options maybe could have been a "Select" component? I feel that you have all the informations to create our custom Select, it's easier to read I think (under public/src/utils.jsx)
- Could we sanitize the elements we send back from the "cubes" API? I know it sounds silly but I just read about this kind of things... 
- What is the use of babel/polyfill?

PR #385, opened 2019-03-06
Hide selected picks option

### LINKED ISSUE #45: hide picks during draft (paper draft simulation)
- hide picks during draft
- review time in between packs

As option in comparison to mtgo like drafts with more time during picks, but without review time between packs.

This would allow for drafts like you're used to at your LGS at FNM or even big tournaments on international competitive level. Has to be an option for sure.

benjaminpwagner:
Add option to hide selected picks during draft to simulate a more realistic draft

Fix #45

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1868,6 +1868,7 @@
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/array-union/-/array-union-1.0.2.tgz",
       "integrity": "sha1-mjRBDk9OPaI96jdb5b5w8kd47Dk=",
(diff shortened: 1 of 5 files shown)


HerveH44:
Hey, don't we want this option to be a "host" option? I had the idea that a host could force players to play in this mode, not that anybody can do that whenever he wants.

Also I don't see in the code the X seconds of time between packs to check your cards.

ZeldaZach:
This was a first attempt, but we can always look at setting it up that way @HerveH44. I'll open another ticket.

PR #391, opened 2019-03-06
Minor fixes

### LINKED ISSUE #387: Use better mana symbols
[symbology.zip](https://github.com/dr4fters/dr4ft/files/2935030/symbology.zip)

use the mana symbols from that zip file instead of the meh ones we pull from wizards :)

### LINKED ISSUE #389: improve pre-selected default sets
The default pre-selection is `Modern Horizons` from type `draft_innovation` currently.
Therefore I assume that always the most current set gets preselected when users come to the page for the first time?

For the majority of users though, this special sets are not interesting and since mtgjson has so much more of these now a better pre-selection could be to focus on sets from type core, expension and master maybe? The basic sets used at FNM events are the ones people are probably going to look for here as well. :)
The drop down list is is huge and it can be difficult to find the right set.

I know this can be seen as feature as well and is intended behaviour. Asking for opinions.

### LINKED ISSUE #390: Set_type in set list dropdown not capitalized
- capitalize `set_type` value used for grouping of sets in the dropdown list
- convert `_` to normal "spaces" in the set_type value to give it a normal look

benjaminpwagner:
Fix #387
Fix #389 
Fix #390


public/src/lobby/Set.jsx
```diff
@@ -17,8 +17,9 @@ const Set = ({ index, selectedSet, type }) => {
         <option value={code} key={code}>{name}</option>
       );
     });
+    const label = setType.split('_').reduce((label, word) => `${label} ${word.charAt(0).toUpperCase()+word.slice(1).toLowerCase()}`, '')
```
ZeldaZach: I swear I wrote a function to do this already elsewhere

benjaminpwagner: Found it in StartPanel.jsx line 7. 

Should I go ahead and make this a utils function?

ZeldaZach: Yeah it should be a util function

PR #440, opened 2019-04-19
add line break to error msg

tooomm:
prevent forced line break on smaller screens


tooomm:
This is **not** working on the heroku deployment:
bad line break

It needs a different solution. @HerveH44? :)

PR #447, opened 2019-04-21
War of the spark: filter guildgates from normal draft

### LINKED ISSUE #444: WAR has Guildgates?
I did a WAR sealed and got a simic guildgate and 8 planeswalkers.   Maybe the second is allowed but pretty sure the first isn't :)

HerveH44:
Fixes #444 

Also:
- fix a minor npm audit problem
- add some tests for WAR (in test/pool.spec.js)
- use date instead of version to update AllSets.json (@ZeldaZach it seems that you bumped the date but not the version on MTGJson, am i correct? I feel that 4.3.2 had first the date 2019-3-30 and then 2019-04-20)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5828,9 +5828,9 @@
       "integrity": "sha1-mGbfOVECEw449/mWvOtlRDIJwls="
     },
     "js-yaml": {
-      "version": "3.13.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.13.0.tgz",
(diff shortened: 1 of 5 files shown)


artofvs:
Is there a way to filter by collector's number? Since not only guildgates should be removed from packs. See comments on https://github.com/dr4fters/dr4ft/issues/444 for more info.

ZeldaZach:
Yeah it’s a mtgjson issue not draft. Closing.

Issue #99, opened 2018-03-02
Simplify configurations override


HerveH44:
Use config.js in src directory that will merge default configs with particular config file

tooomm:
Do you want to open a WIP-PR for this to keep better track of it and have a reference to the branch and actual changes?

I assume it's this branch? https://github.com/dr4fters/dr4ft/tree/configuration

HerveH44:
I've been thinking of this problem. It's important to be able to add configuration's variables while staying compatible with past configuration usage. 

Our situation is that we have default configurations variables for server and client, and we give the possibility to any user to change what he wants by rewriting the files. 

Actually at the end of "npm install", during the "postinstall" phase, we check if the configurations' files exist, and if they don't we copy the default. So we always have the default versions in "config/config.server.js.default" (same for client configuration file) and the EFFECTIVE configurations' files in "./config.server.js" and "./config.client.js". 

A problem arises when we add a new variable in the default configuration and then a server updates the project. This variable doesn't get copied, because the effective configurations' files already exist.

I can see different solutions but right now I'd do that:
Try to merge at runtime the configurations from both server and client. That means that the client's configuration doesn't get compiled, meaning in the end that all the informations of the config get passed to the client by the Websocket. Which means more traffic, especially of static informations (name of the site ...)

@ZeldaZach @dev-id @tooomm what do you think?

HerveH44:
My idea is impossible yet because of the client config file: there can be React components inside, so it's impossible to have it pushed to frontend ...

ZeldaZach:
My idea is to try and do a "merge" of the files, and then if it fails offer a solution similar to how apt does it (open a shell to investigate, view diff, use yours, use maintainers)

HerveH44:
Do you know any good JS library that could merge files ?

I feel a bit insecured about trying to merge 2 JS files ...

HerveH44:
I don't know if anyone is still using our repository for personal projects. In order to "simplify" the configurations I'd like to rewrite all the configuration using env variables and start from there. 

@ZeldaZach @tooomm do you know anybody using this project ?

That would imply breaking the current configuration style. I really don't want to keep the configuration files as they are now, that really look odd.

ZeldaZach:
Fee free to change and we’ll bump to 1.0.0 soon

HerveH44:
[...]

PR #517, opened 2019-06-19
README: update known sites

tooomm:
remove dr4ft.com
~~add trsbox~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,6 @@ It supports all their features, and many more.
 
 #### Known pages running this code:
  - [dr4ft.info](https://www.dr4ft.info)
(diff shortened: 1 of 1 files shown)


tooomm:
Same for me now, maybe they are updating...
I remember that side back from the old times of drafter, and it looks like @trsbox is hosting the new version now too! ❤️

HerveH44:
Well if there's not any other site running dr4ft code, then we could take some time to do the #99 refactor as we can break compatibility with previous configuration setups !

ZeldaZach:
I’m fine with this

tooomm:
I tried several times now in between without luck.

I removed it again.

PR #573, opened 2019-08-05
[#558] check card's URL

### LINKED ISSUE #558: Allow card image override via special tag
A user requested the ability to hotlink to images that are a-typical. Useful for custom sets.

Instead of directly pulling from SF, can pull from custom path instead.

HerveH44:
Should Closes #558 

Could somebody check with a custom set?

### DIFF
--- a/public/src/game/Cols.jsx
+++ b/public/src/game/Cols.jsx
@@ -5,6 +5,14 @@ import App from "../app";
 import {getZone} from "../cards";
 import {Spaced} from "../utils.jsx";
 
+const isURLScryfall = (url) => {
+  try {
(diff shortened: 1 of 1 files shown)


caelen-feller:
There's an issue with custom set upload unrelated to this PR, will submit issue. Can confirm that if the custom set is saved correctly (which it isn't currently), this will work.

It does not account for flip cards, though, which [currently](https://github.com/dr4fters/dr4ft/blob/5e70053ebf2cfe7526bcdf51a007093a3ecde039/src/make/doSet.js#L105-L107) will still try and use scryfall. If you want to just add a check for a field called `flippedCardURL` then this will be supported.

PR #626, opened 2019-09-23
Better look for cube list entry.

queercat:
Just looks better than a single-line textarea.

### DIFF
--- a/public/src/lobby/CubeList.jsx
+++ b/public/src/lobby/CubeList.jsx
@@ -35,7 +35,8 @@ const CubeList = () => {
       <div>one card per line</div>
       <Textarea 
         placeholder='cube list' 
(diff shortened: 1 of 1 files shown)


tooomm:
Did you test this, @ZeldaZach ?

Looking at the Heroku deployment it seems to breaks somehow:
https://github.com/dr4fters/dr4ft/deployments

ZeldaZach:
Looks fine on master. I deployed with no issues

tooomm:
For me, selecting "cube" makes all site turn blank on the heroku deployment.

tooomm:
@queercat Do you want to look further into this?
I also think displaying just one single card name of all list is not great. Unfortunately, your change made the site break...

PR #726, opened 2019-12-01
save state of chat

### LINKED ISSUE #688: Auto hide chat if game is started with only one player
If a player is the only person in a game room and still starts, he doesn't need the chat as there is nobody else he could talk with. :)

I for myself do sometimes pool building practise for sealed. Alone.
First thing I do after I started the game is to hide the chat - every single time I open a new pool. It just takes up valuable space and is of no use for single player rooms imo.

HerveH44:
closes #688 

Save state of chat (do not override the value to true in our case)

### DIFF
--- a/public/src/router.js
+++ b/public/src/router.js
@@ -26,7 +26,6 @@ function route() {
     );
     App.state.players = [];
     App.send("join", id);
-    App.state.chat = true;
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Reach out on Discord sometime! :)

PR #733, opened 2019-12-02
notify when pack is available

### LINKED ISSUE #722: Add desktop notifications as an alternative to beeping on new packs (and for chats)
Copied over from https://github.com/arxanas/drafts.ninja/issues/7, originally opened by @arxanas 

>It's like beeping but there's not actually a potentially-annoying or -inappropriate sound.

<br>
Also, there are notifications now once the chat is hidden for dr4ft.

It would be awesome, to have real browser based web-notifications for chats and/or new packs. Or when the pick timer runs below 15s for example.

HerveH44:
closes #722 

I use the [Notification API](https://developer.mozilla.org/en-US/docs/Web/API/notification) to ask for permission and then use it to send a notification when the tab is inactive (which is what I would expect, I really don't like to be notified if I'm looking at the tab). 

I deleted the beep implementation which is kind of old.


tooomm:
Wow, that works nice. Pretty cool, I like it a lot! :)

Some comments:
- Can you try styling the notifications a bit?
With `Notification.image` and the dr4ft logo for example.

- Hmm, there might still be a use case for the beep. For example if somebody is still on the page but doing other stuff in the room or is distracted for some other reasons. Or he is on a portable browser which mostly have no desktop notification support on windows. Or something is not configured correctly, or... Just keep it as fallback or alternative?
```
Notifications
[ ] Beep on new packs
[ ] Use desktop notifications
```

- Could the permission request to send notifications be triggered every time the check box is selected instead on first visit only? Currently there is no way to re-trigger the request as I understand it.
If somebody declined the pop up for some reason or revoked the permission... at least in Firefox, it's not that easy to manually grant the permission again.

- This only partly solves #722 as there are no desktop notifications for chat messages yet.
That could be a second stage (with a new issue) though.

ZeldaZach:
I tried testing this with a friend but it didn't seem to work... I granted permissions but no checkbox or anything appeared

tooomm:
Browser and os versions? @ZeldaZach

ZeldaZach:
Mac, Safari and Firefox. Latest for both of us.

tooomm:
There should be support: https://developer.mozilla.org/en-US/docs/Web/API/notification#Browser_compatibility

Anything disabled or blocked on OS or browser level? Does other notifications work on Safari?
Try this page: https://www.bennish.net/web-notifications.html (there is authorize on click @HerveH44!)

HerveH44:
There's still work to do, like ask for notification only when the checkbox is clicked. I don't really like the beep but we can think of it as a replacement for desktop notification...

@ZeldaZach it seems really weird that the notifications didn't work. I'll check a bit more in the new days.

ZeldaZach:
Bug on my end w/ FireFox. Works as expected

HerveH44:
> Could the permission request to send notifications be triggered every time the check box is selected instead on first visit only? Currently there is no way to re-trigger the request as I understand it.

@tooomm actually no, it's by design like this. So I'm don't what to implement as the checkbox would work only the first time. As is, I like my first implementation which shows when you arrive on the website. 

Maybe I could keep the checkbox but add a comment on it?

tooomm:
You are right, the request permission on click works on the linked page also just for the first time. Even if you hit ESC instead properly denying... at least with Firefox.
If I use Edge and hit `X` instead `No`, it re-appears everytime I hit the `Authorize` button. So I think it's a browser thing.

>Maybe I could keep the checkbox but add a comment on it?

Keep which checkbox? The beep one?

HerveH44:
I put back the beep and check notification permission through checkbox

tooomm:
That's very cool! Also the useful hint when web notification are blocked. 👍 
The low res icon can be replaced later, we need a better one for the js warning, too. :)

HerveH44:
good for push?

PR #736, opened 2019-12-03
Foil

### LINKED ISSUE #729: Foil effect only applies in non-column view
Picked foil cards (or foils in opened packs in sealed) should still show the foil effect in `Main Deck`, `Sideboard` and `Junk` zones when the view is changed to `Use column view`.

HerveH44:
closes #729 

I'm not that good in CSS so it may be good to review it. I mostly went modifying CSS properties until I found it satisfying.

I also added a property for eslintrc because it couldn't read the .jsx files

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -46,6 +46,11 @@ module.exports = {
     "settings": {
         "react": {
             "version": "detect"
+        },
+        "import/resolver": {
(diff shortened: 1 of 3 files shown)


ZeldaZach:
Good enough for me. I tried it out and it solves the issue at hand in a fine way.

PR #737, opened 2019-12-03
hide move arrows when single player

### LINKED ISSUE #725: Hide up/down arrows for moving position of people for single players
Hide or grey out/disable the up and down arrows used for positioning players in a game room as long as there is only one player.
In a single player game they are useless. One can click them, but nothing happens.

Alternatively (or even better?) those could always be hidden for the host player himself as he doesn't need them when he stays the only player.
In a multiplayer game he can adjust his own position if needed with positioning the other players.

<br>

Can't test right now how it looks when joining a game as guest as they should only be available for the host.
Are they greyed out already for guests? are they not visible at all? Are they clickable but without functionality?

HerveH44:
closes #725 

I chose not to hide arrows for the host, because it's more code to write and it doesn't feel right to me to move the other players when you want to move yourself.

### DIFF
--- a/public/src/game/PlayersPanel.jsx
+++ b/public/src/game/PlayersPanel.jsx
@@ -114,7 +114,7 @@ const PlayerEntry = ({player, index}) => {
 
   if (isHost) {
     //Move Player
-    if(!didGameStart)
+    if(!didGameStart && length > 1)
       columns.push(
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Seems fine. Solves the issue

PR #738, opened 2019-12-03
Show more than one line/card for cube list

### LINKED ISSUE #656: Show more than one line/card for cube list
Displaying just one single card name of all cube list is not great. 

@queercat Do you want to look further into this?

HerveH44:
closes #656


ZeldaZach:
Nice!

PR #739, opened 2019-12-03
keep GameSettings at the same place

### LINKED ISSUE #357: issues with new box layout
This is related to https://github.com/dr4fters/dr4ft/pull/346 and https://github.com/dr4fters/dr4ft/pull/352

- when there are only fewer people (2 friends doing a sealed for example), there is a gap between `Game` and `Settings` box:
layout1

- the distance between boxes is not identical, maybe the added margin (that `Settings` doesn't overlay anymore) differs from the rest?
layout2

CC: @benjaminpwagner

HerveH44:
closes #357 

I put back the GameSettings at the same place. It seems to work on my PC.

### DIFF
--- a/public/src/game/Game.jsx
+++ b/public/src/game/Game.jsx
@@ -40,8 +40,6 @@ export default class Game extends Component {
   }
 
   render() {
-    const { gametype, gamesubtype } = App.state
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Looks good to me

PR #740, opened 2019-12-05
Chat

### LINKED ISSUE #721: Add a full timestamp with date to the chat
Copied over from https://github.com/arxanas/drafts.ninja/issues/31, originally opened by myself.

>Right now there are just times which lead to confusions because you can't really tell how old messages are and days get easily mixed up:
>dates

<br>

Useful comment from the old repo:
>The best UI is perhaps to add a horizontal rule with the date before each section of messages for a given day. Adding the full timestamp to each line would take up a lot of horizontal space.

HerveH44:
closes #721 

This one was quite interesting, I had to rewrite kind of the whole chat code (but the behavior is the same) and also solved the tricky bug of duplicated messages. 

The messages are now grouped by date and have a header which is the day of the message.

### DIFF
--- a/src/make/doSet.js
+++ b/src/make/doSet.js
@@ -48,8 +48,7 @@ function doSet(rawSet, allCards = {}) {
 }
 
 function doCard({card, cards, rawSetCards, code, set, baseSetSize}) {
-  var { name, number, layout, names, convertedManaCost, colors, types, supertypes, 
-    manaCost, url, scryfallId, side, isAlternative, power, toughness, loyalty } = card;
+  var { name, number, layout, names, convertedManaCost, colors, types, supertypes, manaCost, url, scryfallId, side, isAlternative } = card;
   var rarity = card.rarity.split(" ")[0].toLowerCase();
 
   if (isAlternative) {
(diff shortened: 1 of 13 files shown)


tooomm:
I remember the double text issue... :)

I think the date should have commas in this format? And no leading zero for single digit numbers:
`Thu, Dec 5, 2019`

HerveH44:
I'm using the standard date format of Javascript. That's maybe lazyness talking but I'd prefer to keep it simple unless it really hurts user experience.

tooomm:
Totally agree then, I thought you style it anyway.
Will test the date shift tonigth

tooomm:
The date headline disappears now when I type a first message.

HerveH44:
I pushed a new fix.

src/make/doSet.js
```diff
@@ -128,7 +129,10 @@ function doCard({card, cards, rawSetCards, code, set, baseSetSize}) {
     isDoubleFaced: isDoubleFaced,
     flippedCardURL: flippedCardURL,
     names: names,
-    supertypes: supertypes || []
+    supertypes: supertypes || [],
+    power,
+    toughness,
+    loyalty
```
tooomm: Is that something that should go here? https://github.com/dr4fters/dr4ft/pull/742

PR #742, opened 2019-12-06
implement card text

### LINKED ISSUE #614: Card Text Box
Option to show the card text, instead of the card picture, on the dr4ft client.

### LINKED ISSUE #743: Capitalize "Rarity" and "Color" options in column view headers
- Sort cards by `Rarity`:
rarity

<br>

- Sort cards by `Color`:
color

All other are displayed nicely in the interface, like the options for `Type`.

HerveH44:
closes #614 
closes #743 

It's a first attempt. Tell me what you think about it

### DIFF
--- a/public/src/game/GameSettings.jsx
+++ b/public/src/game/GameSettings.jsx
@@ -57,9 +57,10 @@ const SortCards = () => (
 );
 
 const sizeDisplay = {
-  "Small": "Low",
-  "Normal": "Medium",
-  "Large": "High",
+  "text": "Text",
+  "small": "Low",
+  "normal": "Medium",
+  "large": "High",
 };
 
 const CardsImageQuality = () => (
--- a/public/src/game/GameSettings.jsx
+++ b/public/src/game/GameSettings.jsx
@@ -57,7 +57,7 @@ const SortCards = () => (
 );
 
 const sizeDisplay = {
-  "text": "Text",
+  "text": "Text-Only",
   "small": "Low",
   "normal": "Medium",
   "large": "High",
--- a/public/src/cards.js
+++ b/public/src/cards.js
@@ -462,12 +462,12 @@ function Key(groups, sort) {
 
   case "color":
     keys =
-        ["colorless", "white", "blue", "black", "red", "green", "multicolor"]
+        ["Colorless", "White", "Blue", "Black", "Red", "Green", "Multicolor"]
           .filter(x => keys.indexOf(x) > -1);
     break;
   case "rarity":
     keys =
-        ["mythic", "rare", "uncommon", "common", "basic", "special"]
+        ["Mythic", "Rare", "Uncommon", "Common", "Basic", "Special"]
           .filter(x => keys.indexOf(x) > -1);
     break;
   case "type":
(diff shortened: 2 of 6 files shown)


tooomm:
- Text, P/T, Loyalty is missing
text-only
I assume the Text will take up **a lot** of space for some cards, maybe we need some other solution here.

- P/T and Loyalty should only be displayed if there is a value
- P/T should be displayed below the text value

tooomm:
That's way better now! Looks pretty good, text is fine too.

Other TODOs:
- Capitalize Rarities (on cards itself + header when sorting by rarities in column view)
- Capitalize Colors (header when sorting by color in column view)
- Adventures don't have both costs (other aftermath, split, double faced/sided cards might be affected, too)
- Cards are not always nicely displayed next to each other in one line (non-column view)

ZeldaZach:
Agreed w/ ton on Rarities
Not sure what you're on about w/ colors. It's something that is inferred and should not be in the text.
Adventure/split cards are buggin' out a bit.
Seems the card object can contain a full card, but they are in fact weirdly aligned at times in Pack,Main, and SB

tooomm:
>Not sure what you're on about w/ colors.

Filed a separate issue, as I realized that it's not related to this PR: #743

HerveH44:
The manacost of split or adventures cards are a nightmare (I have to modify how we save the cards) and the layout is also a personal nightmare (as I have only very basic CSS skills). I'll look for the manacost but I'd put the layout problem to another issue

public/src/game/GameSettings.jsx
```diff
@@ -57,9 +57,10 @@ const SortCards = () => (
 );
 
 const sizeDisplay = {
-  "Small": "Low",
-  "Normal": "Medium",
-  "Large": "High",
+  "text": "Text",
```
tooomm: `Text-Only`, `No Image` or simply `None`?

There are not all not quite accurate though with the option label being `Card image quality:`... but better than `Text`?

HerveH44:
I pushed a quick fix for the Text option display. I'll stop here as I supposed the rework of the special cards like split, aftermath, adventures and so on would make a big rewrite of how we parse the cards. Also it needs to be very clear on how we want to handle these specific cards in the "text-only" display. I wouldn't really know how to show them at the moment and it would need to be very specific for each of them...

tooomm:
TODO:
- Capitalize rarity information on cards

If the layout is not perfect that's no deal breaker!
But what do you think, @ZeldaZach... Can we merge this with the known issue of mana cost not being displayed for cards with a special layout (adventures, split, aftermath...)?

ZeldaZach:
It's a start, sure. I can add it and gain feedback if you'd like. The real dealbreaker for me is how cards have varying heights

Screenshot 2019-12-08 22 48 06

HerveH44:
pushed a commit that should also close the new PR #743

HerveH44:
Found a display that should the trick for the placement of Text-Only cards

ZeldaZach:
Great start!

tooomm:
Looks very nice! Well done. ❤️ 

I will add the open issue with incomplete mana cost for cards with special layouts to the follow up ticket (#744).

Issue #752, opened 2019-12-10
Improve text-only display of cards


tooomm:
Known issues/limitations of the current first implementation (#742):

- [ ] Mana cost not being displayed completely for cards with a special layout (adventures, split, aftermath...)
Only one half is displayed as realized in ELD adventures.

- [x] Highlight card names in bold (#759)

- [ ] Highlight card names for alternative title in img tag in bold as well, see https://github.com/dr4fters/dr4ft/pull/759#issuecomment-565200634

- [ ] Increase visual separation between column header and its content.
Right now it's difficult because the card names are highlighted in bold. And column headers are bold, too. Use a gap maybe, or increase the size of the column caption a little?

- [ ] Offer option to disable card preview,
or allow the preview to be text-only as well for people who want/need to avoid pictures

- [ ] indicate card frame in text-only, too as guidance

- Also, see https://github.com/dr4fters/dr4ft/issues/769

- Discussion:
Only having the basic type of a card might not be enough in the long run.
As there are many important themes around types, like currently Knights. Artifacts, Legendary, or the Enchantment type of some creatures used to be important, too.
Dr4ft in text-only mode would be way less useful if you can't check those as well.

HerveH44:
For the first bullet point, it needs first to be clear what should be done for all the special cards (aftermath, meld ...) for manacost. Do we want to show the double manacost always or only for specific layout?

tooomm:
My first thought is always.
At least for special layouts on one card site. 
For which Layout wouldn’t it be useful there? For example split/aftermath/fuse, adventure, flip (kamigawa)... they all have more than one mana cost information on one card face. And they should all be shown. 

Special cards like any dfc only have mana costs on one side too I guess.
And if not, just the ones on the currently displayed side matter.

HerveH44:
won't fix

tooomm:
Still want to highlight two points from this issue which are lacking imo and are more important ones in case somebody wants to pick this up:
- Mana cost **(and text!)** not being displayed completely for cards with a special layout (adventures, split, aftermath...)
Only first half is shown as realized in e.g. ELD adventures or GRN split cards.

- Offer an option to disable the highlighted card preview for people who want/need to avoid pictures (e.g. bandwidth reasons or space issues on small screens)

tooomm:
I keep adding to this to keep track of things:

There are scroll bars once there is too much text:
[...]

PR #759, opened 2019-12-12
Display card names in bold

tooomm:
Further improvements to text-only mode.
Partially addresses #752

I couldn't make the alternative text for images, which contains the name too, work yet.

names


tooomm:
@HerveH44 can you help me with making the alternative text for images bold, too?

HerveH44:
I don't know if that's possible. maybe read that : https://stackoverflow.com/questions/7101743/can-i-style-an-images-alt-text-with-css

It seems possible if you add maybe the css property "font-weight": bolder on the "img" inside the span?

tooomm:
Nothing I can deal with right now.
Will add that to the issue.

tooomm:
This is ready for merge by the way.

PR #762, opened 2019-12-12
fix RNA pack

### LINKED ISSUE #753: RNA set does not work
error starting a room when using packs of RNA (ravnica allegiance)

HerveH44:
closes #753 

It appears that the capitalization of Colors and Rarity introduced a regression to how RNA and GRN packs are generated. 

This PR solves the problem by checking the correct rarity (Common instead of common) and also adds checks on others colors' types. 

Though everything would need a refactor to use the types from an enum or something. 

@ZeldaZach Does MTGJson holds JSON schema with maybe enums for colors, rarity etc?

### DIFF
--- a/public/src/cards.js
+++ b/public/src/cards.js
@@ -24,8 +24,8 @@ for (let name in Cards)
     name,
     cmc: 0,
     code: "BFZ",
-    color: "colorless",
-    rarity: "basic",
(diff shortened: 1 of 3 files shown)


ZeldaZach:
Unfortunately no ENUMS available at this time.

Issue #744, opened 2019-12-07
Web notification improvements


tooomm:
Follow up to https://github.com/dr4fters/dr4ft/issues/722 (web notifications got implemented in https://github.com/dr4fters/dr4ft/pull/733)
- [x] Use higher resolution "4" icon (#761)
- [x] Improve arrangement and wording of the two options in settings (#766)
- [ ] Further improve settings display, see description text in PR https://github.com/dr4fters/dr4ft/pull/766
<br>

- Maybe chat support is an option, balance pros and cons

HerveH44:
@tooomm could you take care of this one? If I remember correctly you added the icon (or maybe the svg images of the mana).

Also maybe be more explicit in the issue, I don't know what you're expecting to change at the moment.

tooomm:
Yes. I will take care of some parts. 👍
But will be a while probably until I find time.

HerveH44:
>  But will be a while probably until I find time.
No problem, I guess the author of this issue will not accept that ;)

PR #766, opened 2019-12-13
Improve wording for notification settings

tooomm:
Partially addresses #744

The first is somehow a general switch.
The first turns beep notifications on.
The second is only selectable once the first is enabled.
The second disables the beep from the first.
The second enables desktop notifications via the browser (likely to trigger an os sound on notification event).

It's quite difficult to make clear that the options are dependent on each other, and how ticking the first is needed for the second even though the second disables something the first was providing (beep sound) while superseding that with a web notification and a different system sound most probably.

<br>

**This is a first, simple approach to improve the wording for now.**

---

A better solution would be to also stress the dependence from the first, and/or group the two notification settings and separate them a bit from the rest:
- with an indentation:
```
[x] Show chat
[x] Enable notifications on new packs
  [x] Use desktop notifications over beep
[ ] Add picks to sideboard
[ ] Hide your picks
[x] Use column view
```

- and/or via their own section for example:
```
[x] Enable notifications on new packs
[x] Use desktop notifications over beep

[x] Show chat
[ ] Add picks to sideboard
[ ] Hide your picks
[x] Use column view
```

But I've no idea how to implement that. @HerveH44, maybe you?

### DIFF
--- a/public/src/game/GameSettings.jsx
+++ b/public/src/game/GameSettings.jsx
@@ -10,11 +10,11 @@ const GameSettings = () => (
       <span>
         <Checkbox side="left" text="Show chat" link="chat" />
         {!App.state.isSealed &&
-          <Checkbox side="left" text="Beep on new packs" link="beep" />
+          <Checkbox side="left" text="Enable notifications on new packs" link="beep" />
         }
         {!App.state.isSealed &&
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Good enough start, can always improve

PR #776, opened 2019-12-17
Import XML cockatrice based sets

### LINKED ISSUE #559: Create tool to convert Cockatrice.xml -> MTG.json
Most custom sets people want to upload are Cockatrice-xml based. We should either accept them or create a tool to conver them to mtgjson-json format

HerveH44:
closes #559
Work in progress.

@ZeldaZach do you have any sets to test the PR please?


tooomm:
Do the custom, user created sets in the cockatrice wiki help?

HerveH44:
I used it to start but i'd like now a test of a real set

ZeldaZach:
@HerveH44 http://www.planesculptors.net/set/dreamscape as an example. Check "downloads" for an XML

HerveH44:
It's for cockatrice v3: 

<cockatrice_carddatabase version="3">

HerveH44:
Oups! Wrong button :D

HerveH44:
@ZeldaZach I developed the code for cockatrice v4 not v3 as you mentioned in the issue.

ZeldaZach:
Yeah... Might want to try supporting both if we can

HerveH44:
Looking a bit at the file, there's one thing that caught my eyes: the pictures are not on the internet but in a directory linked with it. We can't upload cards like that, so I don't know if that's possible to force people to upload cards with a true picture URL.

HerveH44:
@ZeldaZach thanks for the file. I updated the parser to also take in consideration the v3 format. It wasn't hard as it's more or less the same format. 

Though I ended up rejecting the file you posted because it lacks the "rarity" on every card. This is too important for the draft (as we are waiting for commons, uncos, rares etc to make the booster). 

As is, I'm happy to push this

public/src/lobby/FileUpload.jsx
```diff
@@ -14,16 +14,23 @@ registerPlugin(FilePondPluginFileValidateSize, FilePondPluginFileValidateType);
 
 const FileUpload = () => {
   const [files, setFiles] = useState([]);
+  const [errorMsg, setErrorMsg] = useState("");
   return (
     <fieldset className='fieldset'>
-      <legend className='legend'>Upload Custom Set</legend>
+      <legend className='legend'>Upload Custom Set (Json/XML)</legend>
```
tooomm: Use capital letters? `JSON/XML`

But I think it doesn't need to be in the headline. It could be announced in the description of the box instead?

tooomm:
Thanks for the update. 👍 

However, `Browse` used to be underlined and got recognized as user-interactable link/object. The mouse changed its cursor and one knew that it's a clickable element to open the open file dialog. That is gone now, and it's not as clear to the user how to load a set.

HerveH44:
@ZeldaZach could you tell me when you think you could integrate this one? I'm working now with the refactor. It could be helpful for me to have this PR merged so I integrate it to the refactor.

PR #886, opened 2020-03-19
fix: show roomInfo for public games

HerveH44:
This should fix a bug on how we display roomInfo. Right now there's an error in the code - we don't show public but private games ... 

I added some tests to be sure that it will work


ZeldaZach:
Your tests seem to be failing, Herve

PR #889, opened 2020-03-19
Fix join player

### LINKED ISSUE #848: Split type cards error in cubes
We should allow users to input one half of a split card instead of saying card not found. 

Ex: saying Braisen Borrower is in my cube should work. Not requiring the adventure half too in the name.

### LINKED ISSUE #884: We can't fire a draft because the lobby doesn't show all of the drafters at once
We can use chat, but the lobby isn't holding players correctly.

image

HerveH44:
should maybe close #884 but I may be wrong

closes #848


ZeldaZach:
How's this coming along?

HerveH44:
I'm waiting for dekkerglen to update the backend of cube cobra. Right now his API is not available from another website.

PR #922, opened 2020-04-04
fix display custom images

### LINKED ISSUE #865: Cube isn't available for custom sets
Hi everyone, first and foremost I apologize if the custom sets I added are causing unwanted clutter and/or if they aren't properly named to reference the game they're part of, I wasn't aware that uploading them would've made them permanently available without me being able to delete them. Onto the issue, as the title says, card names in a cube list are only checked with MTG card databases, and thus I'm unable to issue a cube draft with cards from the custom sets I've uploaded. 
Would it be possible to expand the card databases to accommodate custom sets, or better yet, to be able to optionally specify from which set(s) the cards from a cube are being taken from, in order to avoid possible homonyms?

HerveH44:
closes #865

Hi @user6174 could you try upload and check if you can see your cards now?

### DIFF
--- a/frontend/src/game/Cols.jsx
+++ b/frontend/src/game/Cols.jsx
@@ -5,7 +5,7 @@ import App from "../app";
 import {getZone, getZoneDisplayName} from "../cards";
 import {Spaced} from "../utils.jsx";
 
-const isURLScryfall = (url) => {
+export const isURLScryfall = (url) => {
   try {
     return /scryfall\.com/.test(new URL(url).hostname);
   }catch(err) {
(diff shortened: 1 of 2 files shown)


user6174:
@HerveH44 The images work, thank you again, but as I feared in my issue #865 there's some problems with homonyms: when a custom card has the same name as a magic card, the artwork of the latter is displayed

HerveH44:
Could you please be more specific? What custom card is displayed differently? I guess for cube?

user6174:
Yes, I encountered the issue when testing a cube draft. Some cards that share the same name between my set and MtG are "Heaven's Gate", "Python" and "Treasure Map".
The issue doesn't happen in traditional draft (in my second screenshot Treasure Map is correctly displayed).
This is the cube list I've used: https://pastebin.com/eHrsyLHb
2020-04-05-122820_1534x778_scrot
2020-04-05-133742_1734x763_scrot

PR #923, opened 2020-04-04
allow mosts sets to be playable

### LINKED ISSUE #891: Mystery Booster not showing..
Hey there, dont know if you know but i guess mystery booster was a bit too big to show?

Would be a blast if we could draft mystery in these times..

HerveH44:
closes #891 

I deleted the contol on baseSetSize to allow all sets that don't have this value in mtgjson to be playable (and this means a ton). Just check on the PR deployment to see by yourself.

Mystery boosters are available and playble from what I tested.

### DIFF
--- a/backend/data.js
+++ b/backend/data.js
@@ -119,7 +119,7 @@ const getPlayableSets = () => {
     const { type, name, releaseDate } = AllSets[code];
 
     //We do not want to play with these types of set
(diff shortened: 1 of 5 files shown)


ZeldaZach:
Should only include "draft_innovation", "expansion", "funny", and "masters". That should cover everything.

All types can be found at https://mtgjson.com/structures/set/#type

PR #933, opened 2020-04-06
feat: indicate that the user is waiting on another pack

haganbmj:
Small change to indicate when the user is waiting for another user to pass a pack.
Just used the text that MTGO does.

waiting-for-a-player-to-pick-a-card


ZeldaZach:
Thanks for the contribution! I'd recommend showing "Waiting for the next pack..." instead, as I feel it's more clear

haganbmj:
Works for me.

PR #934, opened 2020-04-06
misc: add CleanWebpackPlugin to clear built directory on build

haganbmj:
Purges the contents of the `built` directory whenever a new package is made. This will help to reduce the clutter in that directory, especially during dev watch.


HerveH44:
Thanks for your contribution!

PR #935, opened 2020-04-07
misc: styling tweaks

haganbmj:
Combined these because they're small and didn't want to flood PRs.

* Displayed player count in Players panel.
* Use css to align input checkboxes and radio buttons rather than whitespace.
* Try to expand on what "Bots" checkbox implies.
* Modify visible columns in Players panel based on draft state/type.

The column visibility change should be revisited. It doesn't account for the Kick/Move player admin options, which already display a visual issue without other players. Also not entirely sure that Kick works once a draft has started, that's either a defect or it means that the Kick button should be hidden after the game has started.

### DIFF
--- a/frontend/src/game/PlayersPanel.jsx
+++ b/frontend/src/game/PlayersPanel.jsx
@@ -5,7 +5,7 @@ import App from "../app";
 
 const PlayersPanel = () => (
   <fieldset className='fieldset'>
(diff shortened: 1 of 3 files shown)


haganbmj:
Yeah disregard what I said about kick, it does replace the kicked user with a bot - I might have been missing something on my local to get it to work.

PR #944, opened 2020-04-07
assign lodash shuffle output 

oelarnes:
the function does not modify the input array


ZeldaZach:
Thanks for the fix! Good catch :)

PR #950, opened 2020-04-09
Added pack statistics generation to bot.js.

queercat:
Will be useful when it comes to creating a smarter bot, this could also be moved to its own file and called from wherever for more useful data as well. For now this seems like an okay place.


queercat:
Actually have some fixes to do to this, not sure why I submitted a pull request.

ZeldaZach:
Thanks for the contribution! Just re-open whenever you're ready :)(

PR #958, opened 2020-04-10
reset window position on new pack

### LINKED ISSUE #799: Reset window position uniformly after manual pick in draft
Manually picking a card in draft should constantly reset the window position to the top of the draftable cards with the headline "Pack x - Pick y/z". Currently the view jumps around after picks. It feels a bit random.
After manually picking and getting presented new cards most likely, it makes sense to center the view to the relevant area again to minimize scrolling.

On autopick, I suggest to not change the displayed content - one might be already building his pool and moving the view can result in unintended clicks to wrong areas.

HerveH44:
closes #799 

It's a starting point, let me know what you think about it @ZeldaZach @tooomm

### DIFF
--- a/frontend/src/cards.js
+++ b/frontend/src/cards.js
@@ -104,6 +104,10 @@ let events = {
       pack[name]++;
     }
     App.update();
+    const anchor = document.querySelector("#cardZone");
+    if (anchor) {
+      anchor.scrollIntoView({ behavior: "smooth" });
+    }
     if (App.state.beep) {
       if (App.state.notify) {
         if (document.hidden) {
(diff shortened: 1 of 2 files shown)


tooomm:
Hmm, sometimes it "focuses/scrolls" to `Main Deck` instead the pick area.
Happens mostly when only the bottom part of the pick area is visible due to scrolling down and picking one of the bottom cards.

ZeldaZach:
I don't see that issue, Tom. I think this is fine 🤷

tooomm:
It still happens for me. At first I thought it happens due to my small resolution/monitor. But on my external display I have the same.

But to be clear, it happens only sometimes and only after some weird click interaction with moving picked cards to the sideboard etc.
I found no way to consistently reproduce it. If you have no idea what might cause it we can probably just merge it as the change is nice.

PR #975, opened 2020-04-16
fix refreshing cards lang

HerveH44:
I introduced a regression on working on last PR, now the cards don't refresh when you change the card lang. This PR fixes it.


tooomm:
Thanks for the fix!

PR #976, opened 2020-04-16
Fix multi card same pack

### LINKED ISSUE #969: Multiple foils of the same type in one pack
This might be related/caused to your recent rules updates @taw, can you chime in and check please?

foils1
foils2
foils4

HerveH44:
closes #969 

This is a big change of how we handle cards in the frontend. Until now we were using a global list of cards that we were accessing by name which resulted in the bug depicted in #969 when two different cards with the same name were indistinguishable (when you clicked on one, it would click on both, or they would be both foil when only one should be, or the contrary, none of them would be foil when one should be). This was an issue on how we saved the cards. 

I changed that by introducing a new class "GameState" that handles the state of the different zones' cards. It helped also implement a nice feature which is that when you click refresh (f5) the cards disposition (your main / side cards, number of lands you inserted). I really feel that would be helpful. 

It still needs some testing as it's mostly frontend and I don't know how to test it.

### DIFF
--- a/backend/pool.spec.js
+++ b/backend/pool.spec.js
@@ -3,47 +3,59 @@
 const assert = require("assert");
 const Pool = require("./pool");
 const {range, times, constant} = require("lodash");
-const { getPlayableSets } = require("./data");
+const {getPlayableSets} = require("./data");
+
+const assertPackIsCorrect = (got) => {
+  const cardIds = new Set();
+  let expectedCardsSize = 0;
+  got.forEach(pool => pool.forEach(card => {
+    assert(card.name !== undefined);
+    assert(card.cardId !== undefined);
+    cardIds.add(card.cardId);
+    expectedCardsSize++;
+  }));
+  assert.equal(cardIds.size, expectedCardsSize, "cards should all have a unique ID");
+};
 
 describe("Acceptance tests for Pool class", () => {
   describe("can make a cube pool", () => {
     it("should return a sealed cube pool with length equal to player length", () => {
       const cubeList = times(720, constant("island"));
       const playersLength = 8;
-      const got = Pool.SealedCube({ cubeList, playersLength });
-      assert.equal(playersLength, got.length);
-      got.forEach(pool => pool.forEach(card => assert(card.name != undefined)));
-      got.forEach(pool => pool.forEach(card => assert(card.draftId != undefined)));
+      const playerPoolSize = 90;
+      const got = Pool.SealedCube({cubeList, playersLength, playerPoolSize});
+      assert.equal(got.length, playersLength);
+      assertPackIsCorrect(got);
     });
 
     it("should return a draft cube pool with length equal to player length per playersPack", () => {
       const cubeList = times(720, constant("island"));
       const playersLength = 8;
       const packsNumber = 3;
-      const got = Pool.DraftCube({ cubeList, playersLength, packsNumber });
-      assert.equal(playersLength*packsNumber, got.length);
-      got.forEach(pool => pool.forEach(card => assert(card.name != undefined)));
-      got.forEach(pool => pool.forEach(card => assert(card.draftId != undefined)));
+      const playerPackSize = 15;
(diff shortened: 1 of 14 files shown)


frontend/src/app.js
```diff
@@ -133,6 +136,20 @@ let App = {
...
   },
+  initGameState(id) {
+    const { gameStates } = App.state;
+    const onStateUpdated = (gameState) => {
+      App.save("gameStates", {
+        ...App.state.gameStates,
+        [id]: gameState
+      });
+    };
+    if (!gameStates[id] === id) {
```
keeler, line 147: nit:
```suggestion
    if (gameStates[id] !== id) {
```

backend/pool.spec.js
```diff
@@ -64,6 +70,7 @@ describe("Acceptance tests for Pool class", () => {
       const [got] = Pool.DraftNormal({playersLength: 1, sets: ["TSP"]});
       assert.equal(got.length, 15);
       got.forEach(card => assert(card.name != undefined));
+      got.forEach(card => assert(card.draftId != undefined));
```
keeler, line 73: What about adding assertions that all the draft IDs are different?

frontend/src/gamestate.js
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   *
+   * @param {string} zoneName
+   * @param {array} cards
+   */
+  addToPool(zoneName, cards) {
+    cards
+      .filter((card) => {
+        return !this.get(ZONE_MAIN).map(({draftId}) => draftId).includes(card.draftId)
```
keeler, line 78: nit: Extract method `cardInZone(zoneName, card) => {...}` because DRY.

frontend/src/gamestate.js
```diff
@@ -0,0 +1,173 @@
+import {countBy, findIndex, pullAt, range, remove} from "lodash";
+import _ from "utils/utils";
+
+export const ZONE_MAIN = "main";
+export const ZONE_SIDE = "side";
```
keeler, line 5: nit: Rename `ZONE_SIDEBOARD` for clarity.

frontend/src/gamestate.js
```diff
@@ -0,0 +1,173 @@
...
+
+function Key(groups, sort) {
+  let keys = Object.keys(groups);
+  let arr;
+
+  switch (sort) {
+  case "cmc":
+    arr = [];
+    for (let key in groups)
+      if (parseInt(key) >= 6) {
```
keeler, line 141: What is this case signifying?

keeler, line 141: nit: Pull out the 6 into a variable `const largeCmc = 6`.

frontend/src/gamestate.js
```diff
@@ -0,0 +1,173 @@
...
+    this.#onStateUpdated(this.inner);
+  }
+
+  getSortedZone(zoneName, sort) {
+    const cards = this.get(zoneName);
+    const groups = _.group(cards, sort);
+    for (const key in groups) {
+      _.sort(groups[key], sortLandsBeforeNonLands, "color", "cmc", "name");
+    }
+    return Key(groups, sort);
```
keeler, line 122: Tests for this new code would be awesome, in particular the sorting/grouping.

ZeldaZach:
I'm happy with this change. Tested and appears to be working for foils and such. If there's a problem, we can re-address in the future

tooomm:
Looks like this solves double cards showing up 👍

HerveH44:
Wait it was still not finished !!!!

HerveH44:
Will push new changes and refactors soon. Don't delete this branch

ZeldaZach:
Sounds good, i'll leave it up

tooomm:
Hmm, maybe add `[WIP]` at the beginning of the PR name next time.
(If you use the new "draft" PR feature, there will be no Heroku deployment and nobody can easily test it)

And Zach should only merge once you removed it... 🙈

PR #982, opened 2020-04-20
frontend refactor

HerveH44:
This PR is just to help maintain the frontend. I splitted "frontend/src/cards.js" into different files:

- cardimage.js : is responsible for making cards image link based on scryfall API style or provide fallback to a default URL
- zones.js: is responsible for zones enum
- events.js: is the new name of "cards.js" that only is the list of events of the App
- gamestate.js : is now responsible for the cards state (so we can retrieve the cards placement after a window refresh), the land distribution (the basic lands that we can add to our pool), and the autopick in the pack.


HerveH44:
@ZeldaZach it's good to go :) !

PR #987, opened 2020-04-23
Better separation of date and time in filename

tooomm:
`-` is already used for the separation of year, month and day (e.g. `2020-04-23`)

Will change the suggested filename
from `IKO-draft-2020-04-23-24523`
to `IKO-draft_2020-04-23_24523` for easier reading.

---

As seen above, the time value is still cryptic and needs proper conversation to local user time.

### DIFF
--- a/frontend/src/app.js
+++ b/frontend/src/app.js
@@ -208,8 +208,8 @@ let App = {
   updateGameInfos({type, sets, packsInfo}) {
     const savename = type === "draft" ? sets[0] + "-draft" : type;
     const date = new Date();
-    const currentTime = date.toISOString().slice(0, 10).replace("T", " ") + "-" + date.getTime().toString().slice(-8, -3);
-    const filename = `${savename.replace(/\W/, "-")}-${currentTime}`;
+    const currentTime = date.toISOString().slice(0, 10).replace("T", " ") + "_" + date.getTime().toString().slice(-8, -3);
+    const filename = `${savename.replace(/\W/, "-")}_${currentTime}`;
 
     App.set({
       filename,


HerveH44:
Thanks @tooomm !

frontend/src/app.js
```diff
@@ -208,8 +208,8 @@ let App = {
   updateGameInfos({type, sets, packsInfo}) {
     const savename = type === "draft" ? sets[0] + "-draft" : type;
     const date = new Date();
-    const currentTime = date.toISOString().slice(0, 10).replace("T", " ") + "-" + date.getTime().toString().slice(-8, -3);
-    const filename = `${savename.replace(/\W/, "-")}-${currentTime}`;
+    const currentTime = date.toISOString().slice(0, 10).replace("T", " ") + "_" + date.getTime().toString().slice(-8, -3);
+    const filename = `${savename.replace(/\W/, "-")}_${currentTime}`;
```
tooomm, line 213: `date.toString()` gives the local user time, e.g. `Thu Apr 23 2020 11:34:04 GMT+0200 (Central European Summer Time)`
But I'm not sure if it always has 3 letter code for day and month...
So `date.toString().slice(16,21).replace(":", "-")` might not work consistently.

But this works, probably there is an easier solution. :)
get hour (two digits)
`date.toString().replace(/\D/g, "").slice(6,8)`
and minutes (two digits)
`date.toString().replace(/\D/g, "").slice(8,10)`
and join them with e.g. "-"

@HerveH44 what do you think?

HerveH44: you can read that if you want : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toString

code is executed on the client computer (frontend = client computer, backend = server)

HerveH44: But I'm not sure if it always has 3 letter code for day and month...
-> it always has 3 first letters unless a month or day that I'm missing have only 2 letters?

tooomm: Nice, according to that docs it should consistently be 3 letter codes, I also checked some other source.
`date.toString().slice(16,21).replace(":", "-")` should totally work then! 👍 

>(...) returns a string representation of the Date in the format specified in ECMA-262 which can be summarised as:
>
>    Week day: **3 letter English week day name**, e.g. "Sat"
    space
    Month name: **3 letter English month name**, e.g. "Sep"
    space
    Date: 2 digit day in month, e.g. "01"
    space
    Year: 4 digit year, e.g. "2018"
    space
    Hour: 2 digit hour of day, e.g. "14"
    colon
    Minute: 2 digit minute of hour, e.g. "53"
    colon
    Second: 2 digit second of minute, e.g. "26"
>
> (...)

Thanks for the helpful input - created a PR. #988

PR #990, opened 2020-04-25
ISSUE-919: Add a decadent draft game mode

### LINKED ISSUE #919: Feature Request: Rich Man's Draft
It'd be great to be able to run a Rich Man's Draft (I've also seen it referred to as Decadent Sealed) through the app. 

For anyone who isn't familiar with it, each player starts with a booster box/36 packs then:

- Opens a pack
- Selects a single card
- Discards the rest of the pack
- Opens another pack and repeats

Once all the packs have been opened, you build a deck from your 36 card pool.

keeler:
# Summary

Fix #919 

Adds a "Decadent" game mode to address the request in https://github.com/dr4fters/dr4ft/issues/919. As described in that issue, it's also known as "Rich Man's Draft" or "Decadent Sealed".

This is essentially a draft where each person picks only one card from a pack, discards the rest, and opens a new pack until all packs have been drafted. As such, it's implemented as a draft where if you would pass, a new round is started instead.

image

# Details
* Only works with the "Draft" game type.
* Can choose as low as 36 or as high as 60 packs to draft from.
* All packs must be from the same set.
    * Implemented this way because editing 36+ set selection dropdowns manually is tedious.
* Refactored `frontend/src/utils.jsx` to split out common components into the `frontend/src/components/` directory.
    * Extracted these components: `Checkbox`, `Spaced`, `Select`, `TextArea`.
    * Fixed in TextArea.jsx `App.save("list", ...)` --> `App.save(link, ...)`
    * Changed `utils.jsx` --> `utils.js`
* Added `frontend/test/` for tests of `toTitleCase()` in utils.js.
* Created `SetReplicated` component which is like the `Set` component except it fills the entire set array with the same set instead of handling one index of the array.
  * Open to suggestions for a better name for this component.
  * Factored out `SetChoices` which is reused in `Set` and `SetReplicated`.

### DIFF
--- a/backend/game.js
+++ b/backend/game.js
@@ -25,6 +25,7 @@ module.exports = class Game extends Room {
       round: 0,
       bots: 0,
       sets: sets || [],
+      isDecadent: false,
       secret: uuid.v4(),
       logger: logger.child({ id: gameID })
     });
@@ -35,6 +36,11 @@ module.exports = class Game extends Room {
     case "sealed":
       this.packsInfo = this.sets.join(" / ");
       this.rounds = this.sets.length;
+      break;
+    case "decadent draft":
+      this.packsInfo = this.sets.join(" / ");
+      this.rounds = this.sets.length;
+      this.isDecadent = true;
       break;
     case "cube draft":
       this.packsInfo = `${cube.packs} packs with ${cube.cards} cards from a pool of ${cube.list.length} cards`;
@@ -342,12 +348,18 @@ module.exports = class Game extends Room {
   }
 
   pass(p, pack) {
-    if (!pack.length) {
-      if (!--this.packCount)
-        this.startRound();
-      else
-        this.meta();
+    if (this.isDecadent) {
+      // Decadent starts new round after each pick.
+      this.startRound();
       return;
+    } else {
+      if (!pack.length) {
+        if (!--this.packCount)
+          this.startRound();
+        else
+          this.meta();
+        return;
+      }
     }
 
     const index = this.players.indexOf(p) + this.delta;
@@ -450,7 +462,8 @@ module.exports = class Game extends Room {
       });
       break;
     }
-    case "draft": {
+    case "draft": 
+    case "decadent draft": {
       this.pool = Pool.DraftNormal({
         playersLength: this.players.length,
         sets: this.sets
--- a/backend/game.js
+++ b/backend/game.js
@@ -191,7 +191,7 @@ module.exports = class Game extends Room {
     super.join(sock);
     this.logger.debug(`${sock.name} joined the game`);
 
-    const h = new Human(sock);
+    const h = new Human(sock, this.isDecadent);
     if (h.id === this.hostID) {
       h.isHost = true;
       sock.once("start", this.start.bind(this));
@@ -348,20 +348,14 @@ module.exports = class Game extends Room {
   }
 
   pass(p, pack) {
-    if (this.isDecadent) {
-      // Decadent starts new round after each pick.
-      this.startRound();
+    if (!pack.length) {
+      if (!--this.packCount)
+        this.startRound();
+      else
+        this.meta();
       return;
-    } else {
-      if (!pack.length) {
-        if (!--this.packCount)
-          this.startRound();
-        else
-          this.meta();
-        return;
-      }
     }
-
+    
     const index = this.players.indexOf(p) + this.delta;
     const nextPlayer = this.getNextPlayer(index);
     nextPlayer.getPack(pack);
@@ -529,7 +523,7 @@ module.exports = class Game extends Room {
 
       if (addBots) {
         while (this.players.length < this.seats) {
-          this.players.push(new Bot());
+          this.players.push(new Bot(this.isDecadent));
           this.bots++;
         }
       }
@@ -549,7 +543,7 @@ module.exports = class Game extends Room {
       this.logger.info(`Game ${this.id} started.\n${this.toString()}`);
       Game.broadcastGameInfo();
     } catch(err) {
-      this.logger.error(`Game ${this.id}  encountered an error while starting: ${err.stack} GameState: ${this.toString()}`);
+      this.logger.error(`Game ${this.id} encountered an error while starting: ${err.stack} GameState: ${this.toString()}`);
       this.players.forEach(player => {
         if (!player.isBot) {
           player.exit();
--- a/backend/game.js
+++ b/backend/game.js
@@ -581,7 +581,7 @@ module.exports = class Game extends Room {
 
       if (addBots) {
         while (this.players.length < this.seats) {
-          this.players.push(new Bot(this.isDecadent));
+          this.players.push(new Bot());
           this.bots++;
         }
       }
--- a/backend/game.js
+++ b/backend/game.js
@@ -413,7 +413,7 @@ module.exports = class Game extends Room {
         this.meta();
       return;
     }
-    
+
     const index = this.players.indexOf(p) + this.delta;
     const nextPlayer = this.getNextPlayer(index);
     nextPlayer.getPack(pack);
--- a/backend/game.js
+++ b/backend/game.js
@@ -193,7 +193,63 @@ module.exports = class Game extends Room {
     super.join(sock);
     this.logger.debug(`${sock.name} joined the game`);
 
-    const h = new Human(sock, this.isDecadent);
+    function regularDraftPickDelegate(index) {
+      const pack = this.packs.shift();
+      const card = pack.splice(index, 1)[0];
+  
+      this.draftLog.pack.push( [`--> ${card.name}`].concat(pack.map(x => `    ${x.name}`)) );
+      this.updateDraftStats(this.draftLog.pack[ this.draftLog.pack.length-1 ], this.pool);
+  
+      let pickcard = card.name;
+      if (card.foil === true)
+        pickcard = "*" + pickcard + "*";
+  
+      this.pool.push(card);
+      this.picks.push(pickcard);
+      this.send("add", card);
+  
+      let [next] = this.packs;
+      if (!next)
+        this.time = 0;
+      else
+        this.sendPack(next);
+  
+      this.autopick_index = -1;
+      this.emit("pass", pack);
+    }
+
+    function decadentDraftPickDelegate(index) {
+      const pack = this.packs.shift();
+      const card = pack.splice(index, 1)[0];
+  
+      this.draftLog.pack.push( [`--> ${card.name}`].concat(pack.map(x => `    ${x.name}`)) );
+      this.updateDraftStats(this.draftLog.pack[ this.draftLog.pack.length-1 ], this.pool);
+  
+      let pickcard = card.name;
+      if (card.foil === true)
+        pickcard = "*" + pickcard + "*";
+  
+      this.pool.push(card);
+      this.picks.push(pickcard);
+      this.send("add", card);
+  
+      let [next] = this.packs;
+      if (!next)
+        this.time = 0;
+      else
+        this.sendPack(next);
+      
+      // Discard the rest of the cards in the pack
+      // after one is chosen.
+      pack.length = 0;
+  
+      this.autopick_index = -1;
+      this.emit("pass", pack);
+    }
+    
+    const draftPickDelegate = this.isDecadent ? decadentDraftPickDelegate : regularDraftPickDelegate;
+
+    const h = new Human(sock, draftPickDelegate);
     if (h.id === this.hostID) {
       h.isHost = true;
       sock.once("start", this.start.bind(this));
--- a/backend/game.js
+++ b/backend/game.js
@@ -38,7 +38,9 @@ module.exports = class Game extends Room {
       this.rounds = this.sets.length;
       break;
     case "decadent draft":
-      this.packsInfo = this.sets.join(" / ");
+      // Sets should all be the same and there can be a large number of them.
+      // Compress this info into e.g. "36x IKO" instead of "IKO / IKO / ...".
+      this.packsInfo = `${this.sets.length}x ${this.sets[0]}`;
       this.rounds = this.sets.length;
       this.isDecadent = true;
       break;
--- a/backend/game.js
+++ b/backend/game.js
@@ -574,12 +574,16 @@ module.exports = class Game extends Room {
     this.startRound();
   }
 
+  shouldAddBots() {
+    return this.addBots && !this.isDecadent;
+  }
+
   start({ addBots, useTimer, timerLength, shufflePlayers }) {
     try {
       Object.assign(this, { addBots, useTimer, timerLength, shufflePlayers });
       this.renew();
 
-      if (addBots) {
+      if (this.shouldAddBots()) {
         while (this.players.length < this.seats) {
           this.players.push(new Bot());
           this.bots++;
--- a/backend/human.js
+++ b/backend/human.js
@@ -1,16 +1,17 @@
 const Player = require("./player");
 const util = require("./util");
 const hash = require("./hash");
-const {random} = require("lodash");
+const {random, pullAll} = require("lodash");
 const logger = require("./logger");
 
 module.exports = class extends Player {
-  constructor(sock) {
+  constructor(sock, isDecadent) {
     super({
       isBot: false,
       isConnected: true,
       name: sock.name,
-      id: sock.id
+      id: sock.id,
+      isDecadent: isDecadent
     });
     this.attach(sock);
   }
@@ -130,6 +131,10 @@ module.exports = class extends Player {
       this.time = 0;
     else
       this.sendPack(next);
+
+    if (this.isDecadent) {
+      pullAll(pack, pack);
+    }
 
     this.autopick_index = -1;
     this.emit("pass", pack);
--- a/backend/human.js
+++ b/backend/human.js
@@ -1,18 +1,18 @@
 const Player = require("./player");
 const util = require("./util");
 const hash = require("./hash");
-const {random, pullAll} = require("lodash");
+const {random} = require("lodash");
 const logger = require("./logger");
 
 module.exports = class extends Player {
-  constructor(sock, isDecadent) {
+  constructor(sock, pickDelegate) {
     super({
       isBot: false,
       isConnected: true,
       name: sock.name,
-      id: sock.id,
-      isDecadent: isDecadent
+      id: sock.id
     });
+    this.pickDelegate = pickDelegate.bind(this);
     this.attach(sock);
   }
 
@@ -112,32 +112,7 @@ module.exports = class extends Player {
     this.draftStats.push( { picked, notPicked, pool: namePool } );
   }
   pick(index) {
-    const pack = this.packs.shift();
-    const card = pack.splice(index, 1)[0];
-
-    this.draftLog.pack.push( [`--> ${card.name}`].concat(pack.map(x => `    ${x.name}`)) );
-    this.updateDraftStats(this.draftLog.pack[ this.draftLog.pack.length-1 ], this.pool);
-
-    let pickcard = card.name;
-    if (card.foil === true)
-      pickcard = "*" + pickcard + "*";
-
-    this.pool.push(card);
-    this.picks.push(pickcard);
-    this.send("add", card);
-
-    let [next] = this.packs;
-    if (!next)
-      this.time = 0;
-    else
-      this.sendPack(next);
-
-    if (this.isDecadent) {
-      pullAll(pack, pack);
-    }
-
-    this.autopick_index = -1;
-    this.emit("pass", pack);
+    this.pickDelegate(index);
   }
   pickOnTimeout() {
     let index = this.autopick_index;
--- a/frontend/src/components/TextArea.jsx
+++ b/frontend/src/components/TextArea.jsx
@@ -1,0 +1,19 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import App from "../app";
+
+const TextArea = ({link, ...rest}) => (
+  <textarea
+    style={ {"overflowY": "scroll", "height": "150px"}}
+    onChange=
+      { (e) => { App.save("list", e.currentTarget.value); } }
+    value={App.state[link]}
+    {...rest} />
+);
+
+TextArea.propTypes = {
+  link: PropTypes.string
+};
+
+export default TextArea;
--- a/frontend/src/components/TextArea.jsx
+++ b/frontend/src/components/TextArea.jsx
@@ -5,11 +5,10 @@ import App from "../app";
 
 const TextArea = ({link, ...rest}) => (
   <textarea
-    style={ {"overflowY": "scroll", "height": "150px"}}
-    onChange=
-      { (e) => { App.save(link, e.currentTarget.value); } }
+    onChange={(e) => { App.save(link, e.currentTarget.value); }}
     value={App.state[link]}
-    {...rest} />
+    {...rest}
+  />
 );
 
 TextArea.propTypes = {
--- a/frontend/src/components/TextArea.jsx
+++ b/frontend/src/components/TextArea.jsx
@@ -7,7 +7,7 @@ const TextArea = ({link, ...rest}) => (
   <textarea
     style={ {"overflowY": "scroll", "height": "150px"}}
     onChange=
-      { (e) => { App.save("list", e.currentTarget.value); } }
+      { (e) => { App.save(link, e.currentTarget.value); } }
     value={App.state[link]}
     {...rest} />
 );
--- a/frontend/src/events.js
+++ b/frontend/src/events.js
@@ -110,7 +110,8 @@ const events = {
     let options = {type, seats, title, isPrivate, modernOnly, totalChaos};
 
     switch (gamesubtype) {
-    case "regular": {
+    case "regular":
+    case "decadent": {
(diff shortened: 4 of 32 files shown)


ZeldaZach:
Tried on the beta build, but it kept crashing

keeler:
> Tried on the beta build, but it kept crashing

Thanks for giving it a shot! Can you share what settings you used? Is there any way to get the logs? There don't seem to be any integration/UI tests and I'm not totally sure what the best way to manually test is.

ZeldaZach:
I tested Decadent mode, 36 packs for both Ikoria and Theros Returned, but they both crashed when I tried to start the game. Unfortunately, I don't have a way to grab those logs from the service. Did it work when you ran it locally under the same conditions?

keeler:
Worked for me with Guilds of Ravnica. But I realize just now that I had # players set to 1. Once I adjusted to 8 it crashed!

Looking into it, and converting this to a draft PR.

Thanks for your patience!

tooomm:
Worked for IKO and THB for me. Tried 36 and 60 packs. 8 player.

- We should dispaly "x times set", e.g. `40x IKO` instead saying `IKO / IKO / ...` 40 times in this mode. :)

info

- I like the name `Rich Man's Draft`  more, that name seems to be used more frequently as well

- Played around with the deployment on heroku and ended up with the new game option showing for sealed as well:

decadent-sealed
I could not start the game however.

keeler:
Thanks for the feedback, @tooomm!

* Updated to display "N times SET": https://github.com/dr4fters/dr4ft/pull/990/commits/26668c1435a952ac66ea20fdf22460704404b7fc
* Choosing "Sealed" + "Decadent" was showing the options for a regular draft. Fixed to show a message that the game mode is unavailable. https://github.com/dr4fters/dr4ft/pull/990/commits/e3a722c970d85b0f94872fd1b25e716c1c4640be
image

I think you're right that `Rich Man's Draft` is a bit more common. This is probably a dumb reason but I decided to go with "decadent" because IMHO it looks nicer in the code itself (e.g. `isDecadent` vs `isRichMans` or `<Decadent ... />` vs `<RichMans ... />`). :grin:

If others feel strongly I can change it to `Rich Man's Draft` but otherwise I'll probably leave it as `Decadent Draft`.

frontend/src/components/TextArea.jsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import App from "../app";
+
+const TextArea = ({link, ...rest}) => (
+  <textarea
+    style={ {"overflowY": "scroll", "height": "150px"}}
+    onChange=
+      { (e) => { App.save(link, e.currentTarget.value); } }
```
keeler, line 10: Note that this was `App.save("list", e.currentTarget.value);` which worked with the code as is since the only instance of `TextArea` had `link` set to `"list"` but would be broken if re-used in other places.

tooomm:
Counting the set looks good, way better. :)

>Choosing "Sealed" + "Decadent" was showing the options for a regular draft. Fixed to show a message that the game mode is unavailable. e3a722c

Can we hide the "decadent" option when sealed is selected?
Switching from draft/decadent to sealed should switch to sealed/regular then.
I think that's more convenient for the user and better UX/UI.

<br>

The new "explanation area" is nice. I really like how this PR/feature is coming along! 👍 

About the name, let's see what Zach and Herve say.

backend/human.js
```diff
@@ -131,6 +132,10 @@ module.exports = class extends Player {
     else
       this.sendPack(next);
 
+    if (this.isDecadent) {
+      pullAll(pack, pack);
```
HerveH44, line 136: I'm not a big fan of having conditions on the game mode, especially if we want to prepare for the winston draft which will be also different. I would maybe use a delegate for the "pick" function or at least a part of it. It can be done later in a refactor if we implement the winston draft. 

I think the simplest way to empty an array is just "pack.length = 0" instead of a "pullAll" but I can be wrong.

frontend/src/events.js
```diff
@@ -109,8 +109,18 @@ const events = {
 
     let options = {type, seats, title, isPrivate, modernOnly, totalChaos};
 
+    if (gamesubtype === "decadent") {
```
HerveH44, line 112: Instead of using the "App.state.setsDraft" I would use a specific property to decouple between the regular mode and the decadent one.

frontend/src/lobby/GameOptions.jsx
```diff
@@ -54,6 +57,30 @@ const Sets = ({ sets, type }) => (
     .map((set, i) => <Set type={type} selectedSet={set} index={i} key={i} />)
 );
 
+const Decadent = ({ sets, type }) => (
+  <Fragment>
+    <blockquote className="game-mode-description">
```
HerveH44, line 62: I love it, could you also add some description for the other modes?

frontend/src/lobby/GameOptions.jsx
```diff
@@ -106,4 +133,12 @@ Chaos.propTypes = {
   packsNumber: PropTypes.number
 };
 
+const Unavailable = () => (
```
HerveH44, line 136: Instead of showing that it's not possible to make a decadent sealed, I would block the "sealed" type and hide the decadent mode when you use the type "sealed".

frontend/src/lobby/GameOptions.jsx
tooomm: Thought the same already - it's awesome! 😍 
But that can also be a new PR I guess...

frontend/src/utils.js
```diff
@@ -0,0 +1,21 @@
+const capitalize = (word) => {
```
HerveH44, line 1: Check lodash : https://lodash.com/docs/4.17.15#upperFirst or https://lodash.com/docs/4.17.15#capitalize

HerveH44:
That's really nice! I have couple questions that I formulated in the different comments in the code. They are all related to how to show in frontend this new mode which is draft only (I would tend to make it impossible to select sealed with decadent) and how not to clutter backend with conditionals (if isDecadent...) that I feel would be cumbersome as we implement other modes (winston or rochester). None of these points are blocking as we can proceed with a refactor later on when we'll have feedbacks or better ideas.

frontend/src/utils.js
HerveH44: I mean I think that we can use directly the lodash functions instead of creating our own (and test them)

ZeldaZach:
I'm supper happy with how this is coming along, but I do agree w/ Herve that we should make this more modular. There are many more draft formats I'd like to support in the future, and decoupling them seems important to allow for maintanability

frontend/src/lobby/GameOptions.jsx
keeler: Glad you like it! Done. https://github.com/dr4fters/dr4ft/pull/990/commits/5850b20df108e249facc6d4d1d19f1ad789e196d

Could use some help wordsmithing.

frontend/src/lobby/GameOptions.jsx
keeler: Done. https://github.com/dr4fters/dr4ft/pull/990/commits/2e29400d94b9e8cf5f5f39191619410cbae3cc48

frontend/src/utils.js
keeler: Good suggestion, done. https://github.com/dr4fters/dr4ft/pull/990/commits/8b109b7d4174f1b5aa548ce38f76940f75e97341

Note that I left `toTitleCase()` because of this https://github.com/dr4fters/dr4ft/blob/82f46024e8680d4dac3df7e906a64e3a66efaebd/frontend/src/lobby/Set.jsx#L22

There's [`startCase()`](https://lodash.com/docs/4.17.15#startCase) in lodash which is pretty close to what `toTitleCase()` does but not quite.

frontend/src/events.js
keeler: Good suggestion. I'll add `App.state.setsDecadentDraft` or something similar.

backend/human.js
keeler: Agreed, this is inelegant for sure. I would prefer to refactor this before merging the PR rather than letting it wait.

> I would maybe use a delegate for the "pick" function or at least a part of it.

Could you elaborate on what you mean by this? I'm not sure I understand what you mean by this, could you spell it out a bit more for me? :confused:

tooomm:
@keeler Heroku deployment shows an empty page after loading for some reason.
Does it work locally for you?

backend/human.js
HerveH44: I am referring to the [delegation pattern](https://en.wikipedia.org/wiki/Delegation_pattern). Instead of using a boolean and handle the different strategies inside the pick function of a player, which is not extendable (as you would need to always add another boolean for another mode, and create a huge function pick with a lot of conditional expressions), you can pass to the player object a delegate (pickDelegate or something like that) that will take care of the pick function. Doing so, you move out of the player pick function the true logic and you encapsulate it to other functions. We could make a "regularPickDelegate" which does what is done right now and make a new "decadentPickDelegate" which will take care of the new pick strategy. 

Doing so, your player object will always take one extra argument, a pick delegate, even when we'll implement new pick strategies. The game though needs to choose the right delegate to associate to the player, but it's fine as it knows what type of game it is. 

Instead of delegation, we could also use inheritance (create a DecadentPlayer - I love the name xD) or create a DecadentGame - I would prefer the latter as I feel that it makes more sense to have one class per type of game, as some game mods don't only affect the pick of a player but also how to create a pool, how to share the pool to players etc (I am thinking of the winston draft pool which is a randomization of X boosters, and the rochester draft which reveals the booster to all the players at the same time even though only one player picks a card.

But these considerations are maybe a bit far fetched. I mean that it may be a big deal to implement a delegate or inheritance, I don't know how you feel about it. So I'm ok with keeping the conditionals, or if you want I can try to implement a delegate to your PR or to another branch to show you what I have in mind.

HerveH44:
Oh, also we should not create bots for the decadent draft ! It does not make sense because players don't get passed any pack..

HerveH44:
@tooomm the PR deploy works for me now. Could you check again?

tooomm:
Nope, still the same. @HerveH44 

The page loads and shortly shows the javascript hint, but then only a white page is shown.

ZeldaZach:
Loads fine for me @tooomm try a diff browser

tooomm:
It does work in other browsers, yes. In Firefox 75 it now suddenly doesn't. :(

frontend/src/lobby/GameOptions.jsx
tooomm: Wording seems ok for a first version. 👍 
Don't want to bloat this PR even more with a bunch of suggestions regarding that.

serokellcao:
Some bikeshedding: I really like the name "Decadent" :)

backend/human.js
serokellcao: @HerveH44 do we really need to go so fancy on that?
Instead, what I would propose is to observe the fact that whilst picking is a player's action, it is still the same for every player in every conceivable draft format, so why don't we just make pick function family static and go for an approach like this:

```
const Human = class {
  constructor (pick_fn, game_type = "Solitaire") {
    this.pick = pick_fn;
    this.game_type = game_type;
  }
  static solitarie_pick() {
    return ("I like it when the cards go psh-psh-psh in " + this.game_type);
  }
  static magic_pick() {
    return (this.game_type + ": the Gathering");
  }
};

const game = {
  player1: new Human(Human.magic_pick, "Magic"),
};

console.log(game.player1.pick())
// expected output: "Magic: the Gathering"
```

It will add another constructor argument to the Human class, but it will keep different pick strategies completely decoupled and it's up to the Game class to select the appropriate function.

serokellcao: At least that's what I'm gonna be doing with my Glimpse Draft implementation unless you people will tell me otherwise :)

serokellcao:
Okay, some non-bikeshedding:

Decadent draft is not a subtype of draft, but a type of draft, so you should really have

`( ) Sealed ( ) Draft ( ) Decadent`.

It's trivial to demonstrate that people might want to play modes "cube decadent" or "chaos decadent".

frontend/src/events.js
keeler: https://github.com/dr4fters/dr4ft/pull/990/commits/e6fd5e583f47f9f5ecdc236fb94d156992346acb

backend/human.js
keeler: Thanks for the information, that was quite helpful!

For now, I've added a `pickDelegate` constructor param to `Human` and removed the `isDecadent` flag from `Player`. https://github.com/dr4fters/dr4ft/pull/990/commits/f9bef937b1889b7f1446a2e342730fe11ca4ef71. I didn't add this to `Bot` because I've actually disabled bots (and player shuffling) for decadent draft in https://github.com/dr4fters/dr4ft/pull/990/commits/f9c3eb58518d384725070526bffdf320c411811f.

As you mentioned, there's the potential for a considerable refactor here. While I want to avoid putting in kludgy stuff like what was there before (the `isDecadent` flag strewn everywhere), I also want to avoid scope creep in this PR. There's undoubtedly a cleaner way to do this (looking at you, @serokellcao, and your commit https://github.com/serokellcao/dr4ft/commit/cc4bacecee8dd6f77cdebcdbab5f3d30686238fd) but IMHO this PR is becoming enough on its own. Please let me know if you disagree.

keeler:
> Okay, some non-bikeshedding:
> 
> Decadent draft is not a subtype of draft, but a type of draft, so you should really have
> 
> `( ) Sealed ( ) Draft ( ) Decadent`.
> 
> It's trivial to demonstrate that people might want to play modes "cube decadent" or "chaos decadent".

Interesting! I think you may be right. What do others think?

keeler:
@tooomm FWIW I'm using Firefox 75 and it's working fine for me. (Ubuntu 18.04.3. LTS)

backend/game.js
```diff
@@ -349,7 +413,7 @@ module.exports = class Game extends Room {
         this.meta();
       return;
     }
-
+
```
HerveH44, line 416: I guess these are useless

backend/human.js
serokellcao: Yeah, it's fine, let's get yours to get merged first, and then figure out what to do with the Human's polymorphism in pick function.

backend/game.js
```diff
@@ -516,7 +581,7 @@ module.exports = class Game extends Room {
 
       if (addBots) {
         while (this.players.length < this.seats) {
-          this.players.push(new Bot());
+          this.players.push(new Bot(this.isDecadent));
```
HerveH44, line 584: I though the Bot could not be decadent?

keeler: Good eye. Fixed https://github.com/dr4fters/dr4ft/pull/990/commits/7d2086c0c339812c5306a64d509af96bd853943b

backend/game.js
keeler: Hmm guess the linter isn't configured to flag extra whitespace. Fixed https://github.com/dr4fters/dr4ft/pull/990/commits/629e5779e1ab5dccd35ed5e7b9c54236e26b2fd2

tooomm:
@keeler 

>Any errors in the console or anything?

I'm on Windows 10. From Firefox web console:
```

Error: "Minified React error #152; visit https://reactjs.org/docs/error-decoder.html?invariant=152&args[]=Q for the full message or use the non-minified dev environment for full errors and additional helpful warnings."
    Ti https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    Oo https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    va https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    cu https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    au https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    Za https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    Hl https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    unstable_runWithPriority https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:30
    Bl https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    Hl https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    A https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:30
    onmessage https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:30
vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14:82483
    ea https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    callback https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    fi https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    ia https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    pu https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    unstable_runWithPriority https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:30
    Bl https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    du https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    Za https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    Hl https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    unstable_runWithPriority https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:30
    Bl https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    Hl https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14
    A https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:30
    onmessage https://beta-dr4ft-issue-919-de-rnrqrn.herokuapp.com/vendors~app~7d359b94-5deb42b5b127fad78a0a.js:30
Error: Minified React error #152; visit https://reactjs.org/docs/error-decoder.html?invariant=152&args[]=Q for the full message or use the non-minified dev environment for full errors and additional helpful warnings. vendors~app~7d359b94-5deb42b5b127fad78a0a.js:14:56527
```

From the Firefox browser console:
```
NS_ERROR_FAILURE: Component returned failure code: 0x80004005 (NS_ERROR_FAILURE) [nsIInterfaceRequestor.getInterface] network-response-listener.js:84
```

Something else I can check?

tooomm:
>>Decadent draft is not a subtype of draft, but a type of draft, so you should really have
>>
>>`( ) Sealed ( ) Draft ( ) Decadent`.
>>
>>It's trivial to demonstrate that people might want to play modes "cube decadent" or "chaos decadent".
>
>Interesting! I think you may be right. What do others think?

@serokellcao @keeler 

I think one can easily argument either way.
When you speak about limited, there is only draft and sealed. With several ways of playing each (aka how you end up with the cards for your "limited pool" to build a deck).

While `draft` and `sealed` is a first simple differentiation whether you pick your cards all for yourself (=build your pool) or you are handed a ready-to-go pool.

I think there are subtypes, but in the end it's the way you put or name it.
As long as there is a consistent pattern to it, and people understand the concept it's ok. But we should keep in mind that many players never heard of the more exotic types of playing and may only know "regular draft" and "regular sealed".
That's why I argue to keep draft and sealed as single top options.
Draft --> Regular Draft, Chaos Draft, Cube Draft, Decadent Draft, Rochester Draft, Glimpse Draft, ...
Sealed --> Regular Sealed, Chaos Sealed, Cube Sealed, ...
There are even some we haven't considered so far like 2HG Draft or Team Draft.
If we make all variations of drafts an individual "top level choice" it will get a bit messy I guess, and sealed will be perished.

*Note: Currently, a "Decadent Sealed" (sometimes offered even at GP level as side event!) is also not possible. If one would call the game mode where every player start with a complete booster box (36 booster) as such.*

<br>

Elaborating your idea (It's trivial to demonstrate that people might want to play modes "cube decadent" or "chaos decadent") even further, the result might be an additional "layer" and could look something like this:

||||||
|---|---|---|---|---|
| **Game Type** | Draft | Sealed | | |
| **Card Pool Type** (=Booster Content) | Regular | Chaos | Cube | ... |
| **Picking Type** (specific to draft game type) | Regular | Decadent | Glimpse | ... |
| more? other? | | | | |
That should keep flexibility and the UI somehow clean.

Because Decadent, Glimpse, Rochester etc. are just picking variations for the limited game mode "draft". For either of those the underlying card pool could be one of several options, too.

**IMO we should keep this discussion out of this PR, finish the feature and find the best UI solution to approach further game modes in an issue ticket.**
**I will create one and link you once this is done.**

ZeldaZach:
I like tom's idea of a 3rd category, although it might add some excessive button clicks to the program. I think it's fine as-is as a separate option that appears when draft is selected. Lets roll w/ it, and if there are issues in the future, we can reevaluate this.

keeler:
@tooomm Thanks for your input on the Game Type + Pool Type + Picking Type. Agree that discussion should be moved out of this PR and into an issue.

Thanks also for the details about the error. I have a hunch that this will fix it. https://github.com/dr4fters/dr4ft/pull/990/commits/a182b9123ce81c913e038530447d0b5d9d040d10

Explanation of my hunch:
Following [the error link](https://reactjs.org/docs/error-decoder.html/?invariant=152&args[]=Q) from the logs you shared indicates that the error is happening because nothing is returned from render. The error started appearing for you after I refactored `GameOptions.tsx` to add the game mode descriptions. Since state is preserved, my hunch is that the state you were in when you last loaded the site you had mode "decadent sealed" selected which now won't load because the switch/case in GameOptions was returning nothing in that case. I have updated that switch/case to default to null which will display nothing. You were the one who IIRC originally gave the feedback about hiding the Sealed + Decadent option so I think this is likely. :crossed_fingers: 

If this does not work then this will require more digging. Unfortunately I don't have access to a Windows 10 machine, so I'd rely on your help here. Perhaps using `git bisect` to narrow down the specific commit where this breaks for you would be a good start. Open to other ideas/suggestions.

tooomm:
@keeler Awesome, that did the trick.
Your assumption was totally right. When I could access the page again it showed "Sealed" as selected, with no second option. Switching between sealed and draft and back showed "Sealed/Regular" selected then.

HerveH44:
Well I still see open my points on the delegate functions being in the game.join function. I don't know if @keeler looked at it or not. I would prefer to go back to implement it to the "pick" function of the player like it was before, because the delegate functions theorically should be pure functions, out in their own file so they can be testable. But this can be done in a refactor, now I guess it's time to move on and let the other PR merge this one.

tooomm:
@HerveH44 Maybe keeler wasn't sure because there was no further response in the related discussion? https://github.com/dr4fters/dr4ft/pull/990#pullrequestreview-400921536

serokellcao:
I agree with the third category.
Re — unnecessary cicks, I'd just have a text field w/ autocomplete as an
alternative input.

keeler:
> Well I still see open my points on the delegate functions being in the game.join function. I don't know if @keeler looked at it or not.

@HerveH44 Are you referring to this thread? https://github.com/dr4fters/dr4ft/pull/990#discussion_r415786771

If it is, I did attempt to address this by adding delegate functions for regular and decadent draft picks, though it's arguably not ideal. I didn't feel comfortable just marking it resolved like some of your other feedback which was much more straightforward (e.g. use lodash functions).

If that's not the thread you're talking about I'm having trouble finding what thread you mean, this PR is getting a bit cluttered.

HerveH44:
Hi @keeler, sorry for the misunderstanding. The PR is really amazing and will help us integrating the new mods in the future. Thanks for your contribution.

PR #1020, opened 2020-05-04
[No Merge] Update wording on "Copy deck" button

tooomm:
TEST


ZeldaZach:
This feels implied. Why use more words if not needed?

tooomm:
Sorry, we don't need to merge the change on the label. It's ok as is.
This was mostly a test to see if other heroku deployments are working fine in my setup. But closing the PR would remove the deployment as well. :)

Just leave it around for a while, please.

serokellcao:
> Why use more words if not needed?

Think [quote](https://www.youtube.com/watch?v=_K-L9uhsBLM) "if few word do trick".

PR #1037, opened 2020-05-15
Booster rules version

### LINKED ISSUE #1024: Feature Request: display magic-sealed-data version
I love how on the landing page of dr4ft.info it shows clearly the build deployed + mtgjson version.
I think it would also be cool to show which version of  https://github.com/taw/magic-sealed-data the server is currently using.

I also think it's nice because it provides a nod of thanks to @taw who is maintaining that resource :heart:

HerveH44:
closes #1024 

added a fix on download of allsets too


tooomm:
I very much like the cut-down in digits @ZeldaZach

Issue #1034, opened 2020-05-15
Dropdown menus revert back to default options


7stringplayer:
So i tried to change the packs and they end up going back to Ikoria

tooomm:
Interesting. Just tested this... and you are right. They dropdown flickers back to IKO when you create the sealed room - barely noticable. I think that's new.
But in the room description you see the right ones listed. When I started the game, the right packs appeared as well.

7stringplayer:
Mine did not they all went to Ikoria not the set i selected

tooomm:
When do they change back to ikoria?
Which site do you use and does it currently still happen?

7stringplayer:
Almost immediately they flock back to Ikoria.  I create the room and it says M20 and 5 IKO.  I use Dr4ft.info

HerveH44:
@keeler It may be linked to your PR. Do you have some time to check if they are linked?

tooomm:
Note: This also happens for draft. It is not a sealed only issue.

Happens in Chromium and Firefox, sometimes immediately, sometimes a bit delayed or once you click the create button.

ZeldaZach:
Server ran out of storage. I cleaned it up just now. Feel free to try again

tooomm:
Unchanged :(

7stringplayer:
Mine seemed to work in Chrome.... no issues now

tooomm:
Works here now, too.

tooomm:
Looks like @ZeldaZach reverted dr4ft.info to a commit without the `Decadent` option.

If I look at https://beta-dr4ft.herokuapp.com/ the issue appears again and seem to be related to that change. Can you confirm @7stringplayer?
Thinking about if further... it looks like the pack selection gets reverted to default instead "remembering" the selected value and present that to the user on next visit.
Because changing the set selection, creating a room, closing the page and going to the page again presents 3xIKO again instead my last selection. It used to remember a players last selection.

CC: @keeler

7stringplayer:
Yes on that site it  is doing what it did yesterday.  The main site is still working

keeler:
Likely related to my recent change, I can take a look this weekend.

I'm a little confused, can I get a concise description of the expected behavior vs what's happening instead?

Is it one of these?
1. In the lobby, select a set other than Ikoria. The dropdown immediately switches back to Ikoria rather than updating to the new set.
[...]

PR #1039, opened 2020-05-17
ISSUE-1034: Fix sets flickering and not remembered

keeler:
# Summary

Fix issue https://github.com/dr4fters/dr4ft/issues/1034 introduced by https://github.com/dr4fters/dr4ft/pull/990 where:

1. The set dropdowns momentarily flicker back to the default set (currently Ikoria) when creating a game room if a set other than the default was chosen.
2. When returning to the Lobby page, rather than remembering set selections all the sets were reset to the default.


tooomm:
Thanks for the quick investigation and fix!!

PR #1040, opened 2020-05-17
fix hash lenght

tooomm:
Truncated GitHub hashes for commits are 7 digits, not 8.

### DIFF
--- a/backend/data.js
+++ b/backend/data.js
@@ -250,7 +250,7 @@ const getBoosterRulesVersion = () => {
       return "";
     }
   }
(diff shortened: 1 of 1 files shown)


ZeldaZach:
There's no issue with having 8 symbols. More information is better here imo.

tooomm:
What kind of information do we miss when we use the default number of 7 digits? 🤔 

Also, that's what GitHub has on his site everywhere and users are used to.
Even we use 7 digits for the version hash of the dr4ft project, it makes sense to use the same here, too.
commit-hash

PR #1042, opened 2020-05-17
cube options ui improvements

tooomm:
Use `[explanation]: [value]` as we do in all interface.
I also changed the display a bit and used a more detailed description for the options.

### DIFF
--- a/frontend/src/lobby/GameOptions.jsx
+++ b/frontend/src/lobby/GameOptions.jsx
@@ -147,12 +147,14 @@ const CubeSealedOptions = () => (
 
 const CubeOptions = () => (
   <div>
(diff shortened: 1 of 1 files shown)


tooomm:
Jep, nice catch. But nothing to do with this PR I guess.
Try to start a normal draft game in the heroku environment for this PR. Probably you'll have the same issues there. Once I selected other sets and was able to start a game, the cube worked as well for me.

Looks like the problem described in https://github.com/dr4fters/dr4ft/issues/1034#issuecomment-629811273 could be even bigger for new visitors. @keeler 
I assume that triggers here because we are first time users for each new heroku deployment.

PR #1049, opened 2020-05-18
Do not always add a colorless card

HerveH44:
## Fixes #
No issue yet :D

This PR tends to solve the colorless card absurd probability representation that I've noticed in boosters when we generate a balanced by color pack.  
I just delete the auto include colorless card.


ZeldaZach:
Good catch

PR #1051, opened 2020-05-18
Improve draftlog generation

### LINKED ISSUE #1035: Client id exported to draft log as event id
Hi, I've been familiarizing myself with this great project to customize it for a league I run.

I had some confusion about App.state.id vs the room id and it looks like I am not the only one!

getLog in events.js write App.state.id to your draftlog as 'Event #', but that value is a uid generated by the client to persist identity between sessions. Looking over logs, it is the case that this is a player identifier not an event identifier.

```
  getLog() {
    const {id, log, players, self, sets, gamesubtype, filename} = App.state;
    const isCube = /cube/.test(gamesubtype);
    const date = new Date().toISOString().slice(0, -5).replace(/-/g, "").replace(/:/g, "").replace("T", "_");
    const data = [
      `Event #: ${id}`,
      `Time: ${date}`,
      "Players:"
    ];
```

In general, the game id and title do not appear to be persisted to App.state nor does title appear to be accessible to the client at all.  I may be able to submit a pull for this issue in the coming days.

oelarnes:
- Fixes #1035 

* Add gameId to App.state, and push gameId and title to app from backend on greet
* Use gameId as "Event #" in draftlog
* Use index + 1 for "Pack 1 Pick 1"

## Screenshots

image


ZeldaZach:
Thanks for the contribution! :)

PR #1060, opened 2020-05-21
fix booster rules version

HerveH44:
## Linked tickets
- Fixes #1040

## Explanation of the issue
The booster rules version are always redownloaded because we save only the first 7 digits of the hash and check with the entire commit hash

I moved the PR intent (substring 7 first digits of the hash) to frontend. I also added some logs for completeness


tooomm:
That's how it should be done of course... 🙈 
Thanks for getting it right!

Issue #1155, opened 2020-07-22
Bug: Sorting and image quality selector messed up in game


tooomm:
**Webpage**: dr4ft.info

#1136 introduced a regression.

## Detailed Description

There used to be a nice horizontal radio button selection table.

## Further Information

Before #1136:
![Untitled](https://user-images.githubusercontent.com/9874850/88186104-0fe1a300-cc35-11ea-9cbe-b7f2435f4dbe.png)

After #1136 it loos like this:
(Don't be confused by the additional options, the above is from a Sealed game, while this is from a Draft with more settings)
![Untitled2](https://user-images.githubusercontent.com/9874850/88186110-11ab6680-cc35-11ea-9f0a-77670bb59175.png)

mixmix:
Is the desired state to revert back to the original state, or was there a new design e should ask for?

tooomm:
It seems obvious to go back to the initial look/behavior.
If you have a different idea feel free to share it here!

mixmix:
I'll _draft_ something up. Working on this now

mixmix:
here you go @ZeldaZach / @tooomm 

https://github.com/dr4fters/dr4ft/pull/1248

mixmix:
Bump @tooomm @ZeldaZach

I think the failing tests are just linting

PR #1248, opened 2020-10-05
fix settings radio buttons, add scss support

mixmix:
## Linked tickets
- Fixes https://github.com/dr4fters/dr4ft/issues/1155

## Explanation of the issue

Settings radios were not visible

- change settings to be highlighted buttons
- add scss support
     - made making this change easier
     - should make future design changes better scoped, easier to maintain

## Screenshots

image

### DIFF
--- a/frontend/src/game/GameSettings.jsx
+++ b/frontend/src/game/GameSettings.jsx
@@ -2,9 +2,10 @@ import React from "react";
 
 import App from "../app";
 import Checkbox from "../components/Checkbox";
+import './GameSettings.scss';
 
 const GameSettings = () => (
-  <div className='game-settings'>
+  <div className='Game-Settings'>
     <fieldset className='fieldset'>
       <legend className='legend game-legend'>Settings</legend>
       <span>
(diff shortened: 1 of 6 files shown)


mixmix:
I am well aware that I fixed a bug AND added a feature (scss) support.
I chose to do this because the current styling support (css) is monolithic and hard to engage with - basically because CSS gets really out of control if you don't pin it down clearly, as you can have styles having wild unintended side effects across the app.

What I've added will make it more likely for me (and devs like me) to engage with improving the UI

If people don't like scss, they can still write css, but this allows them to at least add it in a file associated with the component

tooomm:
Thanks for the fix and opening a PR right away @mixmix!

However, I can't check this. Hope @ZeldaZach gets to it.

<br>

Your solution to use buttons over the initial radio buttons is fine and a big improvement over the current broken styling! Thanks for providing a screenshot as our Heroku Review App didn't trigger to build a live version of the change.

May I ask why you chose to not align the buttons more to the right?

mixmix:
Thanks for having a look @tooomm

Re buttons being right aligned, I'm not 100% sure what you mean? I went for indented and left aligned because that seemed tū be the pattern followed elsewhere. If you want to clarify what you mean with a sketch happy to adjust.

Also if you'd like to see what this looks liked in practice happy to deploy this to an instance I'm running

mixmix:
I see i gun you meant why did I indent? Yeah given these are not sub options like elsewhere I think the indent of be removed. I'll change that

tooomm:
Haha, yeah... I got confused myself. Of course I meant left aligned.
I asked because it looks like it doesn't align with anything else currently regarding your screen:
indent

I'm not saying I dislike it per se, but then the card langauge selector needs the same adjustments at least?
Anyway, the gap feels kinda big. Maybe you can test both and see what looks and feels nicer for the panel overall.

Thanks for the kind offer to link to your instance. But here the screen is good enough for me I think. :)

mixmix:
image

yeah that looks better (cc @tooomm )

ZeldaZach (approved):
I'm very happy with this change! Thank you so much for your support of our project :)

tooomm:
Thanks for having a second look! @mixmix

mixmix:
I hope this is the first of many bits of support I can offer!
Thanks for the reviews and fast merge <3

PR #1302, opened 2020-12-03
Add support for specifying particular card versions in a cube

nex3:
A cube card can now be written as "Abrade (CMR)" or "Abrade (CMR 659)" to specify the exact image that should be displayed. To enable this, the cubable card database has been expanded to include a map from card names to set codes to version numbers, as well as a "default" version for each card that's based on the same logic that's historically been used to select images.

If this approach looks good, I'd like to follow it up with an additional PR that loads card versions from CubeCobra, so that the versions drafted here match the versions specified in the cube list itself.

## Screenshots

Here's an image of a cube draft showing off different images for the same card:

image

### DIFF
--- a/backend/data.js
+++ b/backend/data.js
@@ -2,7 +2,7 @@ const fs = require("fs");
 const _ = require("lodash");
 const path = require("path");
 const readFile = (path) => JSON.parse(fs.readFileSync(path, "UTF-8"));
-const { keyCardsUuidByName, keyCardsUuidByNumber, groupCardsBySet, groupCardsByName } = require("./import/keyCards");
+const { keyCardsUuidByNumber, groupCardsBySet, groupCardsByName } = require("./import/keyCards");
 
 const DATA_DIR = "data";
 const DRAFT_STATS_DIR = "draftStats";
@@ -110,7 +110,7 @@ const getCubableCardUuidByName = (cardName) => {
   if (card.number && byNumber[card.number]) return byNumber[card.number];
 
   return byNumber[Object.keys(byNumber).sort()[0]];
-}
+};
 
 const getCubableCardByName = (cardName) => {
   return getCardByUuid(getCubableCardUuidByName(cardName));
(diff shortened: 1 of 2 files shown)


backend/data.js
```diff
@@ -80,11 +81,39 @@ const getCardByUuid = (uuid) => {
...
 };
 
-const getCubableCardByName = (cardName) => {
+const parseCubableCardName = (cardName) => {
+  // Cube cards can be written as:
+  //
+  // * "Abrade" (just card name)
+  // * "Abrade (CMR)" (card name + set code)
+  // * "Abrade (CMR 410)" (card name + set code + number within set)
+  const match = cardName.match(/^(.*?)(?: +\(([a-z0-9]+)(?: +([a-z0-9]+))?\))? *$/);
```
ZeldaZach, line 90: Looks like your regex doesn't match what it should be matching. you forgot about capitals in your set code and card number. The fix would be `[A-Za-z0-9]+`, but you can shorthand that to simply `\w+`, which is what i did in my replacement.

```suggestion
  const match = cardName.match(/^(.*?)(?: +\((\w+)(?: +(\w+))?\))? *$/);
```

nex3: The card names are all lowercased before they're passed here, but you're right that `\w+` is more obvious and more terse.

HerveH44:
hi @nex3 , sorry I didn't see at all the PR. I just recreated the deployed app for this PR and I quickly tried a cube with the "8ed" cube from cubeCobra and the app couldn't find the cards. I did not dig more but I guess there's some problem to recognize cards ?

HerveH44:
On my computer it seems that there's a problem with the sets not being parsed. I'll take a look deeper later. I guess it may be linked with migratiion to MTG v5. A quick merge of dr4ft/master should solve the issue.

HerveH44 (approved):
Well done!

LotusWalker:
This will be quite a big improvement, thanks! @nex3 Also think your idea for a follow up PR which pulls the specific card versions we have on CubeCobra into Dr4ft will be an awesome upgrade to the program.

mixmix:
This is choice! Great work @nex3

PR #1308, opened 2020-12-07
actually use different versions in tests workflow

ebbit1q:
## Explanation of the issue
The current tests are run with the names lts and latest but are actually just run twice on version 12, the check-latest tag does nothing.

instead run them on version 12 and 14, travis used 15 which seems unavailable.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,9 +18,9 @@ jobs:
 
     strategy:
       matrix:
-        version:
-          - latest
-          - lts
+        node_version:
+          - 12
+          - 14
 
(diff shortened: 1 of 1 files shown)


tooomm:
@ebbit1q FYI:

>The current tests are run with the names lts and latest but are actually just run twice on version 12

Travis used to have those... GH Actions don't support `lts` or similar aliases just yet.
There is a ticket in their tracker already - number 26: https://github.com/actions/setup-node/issues

>the check-latest tag does nothing

That tag is only supported in the "v2-beta" action, not the default v1:
https://github.com/actions/setup-node#usage

PR #1326, opened 2020-12-10
Refactor state mgmt

### LINKED ISSUE #1312: Feature Request: Refactor frontend state management
## Detailed Description
State management is a mess. Few things I'd like to touch to simplify:

    Separate state of Game Creation (all the options, that are actually in "App.state") from State of the Game played (which is a bit everywhere, there's a start in "App.state.game".
    GameState is doing too many things: Simplify with one component in charge of saving the cards in one place, so we would have 3 subcomponents, each handling one place (MAIN, SIDE, JUNK). I doubt that we need to save the state of Pack as it's sent by the server at connection. We need also a component to save autopick / burn id of current pick. Then GameState could just be a HoC that just wrap all of it. It should also be renamed because it clashes with App.state, or ... merge App.state into a new subcomponent of GameState that handles game creation options. That would indicate future devs where they should put/get the data in future features.

## Further Information
We could use Redux as a clean and standard way to handle state management across the app. Anybody has a better suggestion ?

HerveH44:
## Linked tickets
- Fixes #1312 

## Explanation of the issue
The state management is a big mess and we should adopt a simpler and more maintainable way of handling frontend app state

I'm starting small to allow incremental changes. I just introduced Redux store capability and connect the store to StartPanel controls.
I added migration capability to make it invisible to the user that the data is being migrated to the new store.


ZeldaZach:
Any reason the beta deployments failed?

ZeldaZach (approved):
Tested locally on multiple browsers and accounts, saw no issue. I agree with moving things to more concrete classes instead of spewed across the code base. Thanks as always :)

HerveH44:
@ZeldaZach I don't know if you noticed but this is just a first step. Not all the data is being handled by redux stores.

HerveH44:
Actually, I just read : https://dev.to/selbekk/persisting-your-react-state-in-9-lines-of-code-9go

I need to think about it a bit, because we may not need actual global state, I need to check again but if we can get along with just local state + persistance we could just be happier. @ZeldaZach I don't know who could have an expert POV on this subject, if you know someone let me know.

ZeldaZach:
cc @lrawicz

mixmix:
@HerveH44 I'm keen on redux. In particular being able to write tests around certain bits of logic (like what should happen when you burn a card you've previously picked?)

ITO using local storage and hooks isn't that just global state but somewhere else and harder to test?

Very excited to see this as part of the app, I think it will make unbeatable the so and contributing a lot easier. Happy to help out with this where possible

HerveH44:
@mixmix thanks, I'll be glad to share thoughts on Redux. I've never put frontend under test, so I would love to know more about it. This PR intent is to be super small and just starts the migration to Redux. I already started with another part of state but it requires a lot of changes.

@ZeldaZach @mixmix if you think this is fine, we can merge this part and move towards another part of state refactor

frontend/src/state/migration/index.js
```diff
@@ -0,0 +1,31 @@
+const readAndDeleteFromLocalStorage = (key, defaultValue) => {
+  const val = localStorage[key];
```
mixmix, line 2: Should probably use localStorage.setItem

getItem and removeItem?

I guess that protects against accidentally mutating wrong things on the raw object?

frontend/src/state/start-controls.js
```diff
@@ -0,0 +1,41 @@
...
+      state.useTimer = !state.useTimer;
+    },
+    setTimerLength: (state, action) => {
+      if (timers.includes(action.payload)) {
+        state.timerLength = action.payload;
+      }
+    }
+  }
+}
+);
```
mixmix, line 32: These reducers are what we could easily write tests over. Haven't seen this createSlice  thing before but we just need a way to reach in and check those reduce functions. These ones are simple, but would be good to get into habit of putting tests over ones which have multiple paths over them.

I would say we should have a testing pattern in place before merging this. Would set us up well. Happy to help

frontend/src/components/Checkbox.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 
 import App from "../app";
 
-const Checkbox = ({link, text, side, onChange, ...rest}) => (
+const Checkbox = ({link, text, side, onChange, value, ...rest}) => (
```
mixmix, line 6: Should add proptyes for value?

frontend/src/game/StartPanel.jsx
```diff
@@ -40,21 +42,41 @@ const StartControls = () => {
 };
 
 const Options = () => {
-  const {useTimer} = App.state;
-  const timers = ["Fast", "Moderate", "Slow", "Leisurely"];
```
mixmix, line 44: Reviewing on my phone but I think you still need this

mixmix (changes requested):
There are modules I've not used here but the code looks reasonable.

I left a few minor comments. My major request before merging would be a single test (as a example for those later) for a reducer.

This is gonna be a great improvement!

frontend/src/game/StartPanel.jsx
HerveH44: It's in the initialState in the reducer.

mixmix:
I'm keen to see redux put in place. The app starts is kinda stressful currently.

This PR seems like a great start.
Who would like to help get this updated and merged? I'm up for figuring stuff out if noone else is (eg how to write tests with this current thing).

Do we also need to consider how any recent changes to react have influenced this work?

PR #1351, opened 2020-12-31
Glimpse draft (UI)

mixmix:
This branch changes:
 - adds SVGs for glimpse draft indicating pick/ burn
 - changes class draft to have same pick svg
 - refactors "Card" components out of `Grid.jsx`
     - `CardBase.jsx` is a raw display component, and has some improvements for readability of card while any image is loading
     - `CardDefault.jsx` the functionality of the classic draft. this just wraps CardBase with a listener for clicks
     - `CardGlimpse.jsx` adds new functionality allowing picking + burning
 - refactors `Grid.jsx` styles to use css-grid
     - span was being abused a lot ):
     - conserved flexible layout while allowing easy handling of spacing
 - a new webpack loader:
     - this makes it so you can reference media files in scss files using easy to understand paths

I made some of these changes because I didn't want to awkwardly weave a bunch of logic through the existing components and make it a total mess. My intention here was to put a little bit of energy in to make the code easier to read, maintain, and extent (both js + html + css)


LotusWalker:
Awesome, thx so much for your work on this! A flame for the “burn card” symbol would be pretty cool. Happy New Year.

mixmix:
Yeah the ForkAwesome flame isn't that clear.
I think if the language around burning is around burning in the settings then it makes sense to be a flame.
I wonder if we need to think annoy how to educate people about glimpse, or can simply link to an explained somewhere

LotusWalker:
Yeah, I think most ppl understand what burning is when drafting. Definitely if they’re choosing the Glimpse draft option. But I agree that a tooltip hint or link makes sense too.

mixmix:
image

here's what it currently looks like
on the left is a card which has already been "picked"
on the right is a card that's being hovered over, revealing the pick + burn options

mixmix:
one moment, I've broken double sided cards. working on it

mixmix:
READY

Issue #769, opened 2019-12-15
Card info for the image alt attribute is badly styled


tooomm:
- Card text starts in the same line as type and rarity
- Power and toughness is not separated with a dash
- Power and toughness starts in the same line as text

![alt text](https://user-images.githubusercontent.com/9874850/70861483-ea6a1980-1f2e-11ea-8610-510dbbbc8230.png)

And as mentioned in #752 already, the name should be bold there, too.

HerveH44:
won't fix

tooomm:
Probably difficult to change (if at all) and not worth it as it rarely ever happens.

Issue #768, opened 2019-12-15
Bad layout when card image can't be loaded and alt text is shown


tooomm:
It breakes the layout if one card image can't be loaded:
![alt text](https://user-images.githubusercontent.com/9874850/70861229-3d41d200-1f2b-11ea-994a-10af22b7af81.png)

This can easily be reproduced as the image for the common card "Ulvenwald Captive" is currently not available at scryfall (but reported), and thus can't be loaded.

HerveH44:
What is your OS and browser?

The card "Ulvenwald Captive" is back in scryfall I guess because I see it currently.

tooomm:
Yeah, they fix reported mistakes very fast normally.

I use Firefox 70 on Windows 10.

HerveH44:
won't fix

HerveH44:
won't fix

PR #1369, opened 2021-01-05
fix styling of text view of cards

mixmix:


mixmix:
Added some simple styling to text version of cards so that they can be made more sense (while still readable) 



Forced text to stay within border of card size for small cards

tooomm:
Related reports and input: https://github.com/dr4fters/dr4ft/issues/752, https://github.com/dr4fters/dr4ft/issues/768 and https://github.com/dr4fters/dr4ft/issues/769

Thanks for taking the time to improve this and applying some magic @mixmix! :)

How does it handle split cards and their mana cost?

mixmix:
@tooomm I'm not sure honestly. Fwiw I've only really moved around what was already there.

In doing this I've definitely realized that text view of cards isn't perfect. Eg double sided cards are not supported

PR #1371, opened 2021-01-06
rename pick + burn vars as types of "selection"

mixmix:
I've been reading this glimpse logic to see how I can support, and noticed that I was quite confused by the names.
This PR is a trial in aligning the language to be more consistent.

I know that renaming variables can be contentious, so if this does not serve or help, then I'm happy to change this around or close it.

| old language | new language |
|---|---|
| **autopicks** | selected.picks |
| **burns**     | selected.burns |
| isPickReady | isSelectionReady |
| pick      | confirmSelection |
| pickonTimeout | handleTimout |
| pickState (event) | setSelected |

In general:
- cards we're marking are "selected"
- we can select for "pick" (keep) or "burn" (destroy)
- autopick is more a concept about what happens on timeout which is a special case, so we remove the auto prefix most places

### DIFF
--- a/backend/bot.js
+++ b/backend/bot.js
@@ -2,7 +2,7 @@ const {sample, pull, times} = require("lodash");
 const Player = require("./player");
 const logger = require("./logger");
 
-module.exports = class extends Player {
+module.exports = class Bot extends Player {
   constructor(picksPerPack, burnsPerPack, gameId) {
     super({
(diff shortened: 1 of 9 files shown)


mixmix:
Thanks for merging this!
Is gonna make working on the rest of the bugs a little bit easier

PR #1382, opened 2021-01-13
glimpse-draft: UX + bugfix

mixmix:
This PR:
- changes the icons to use existing ioninc icon font!
- updates ionic icon (a little)
- changes the design of what the "selected" cards look like
- fixes a tiny bug

Now looks like this : 
image


frontend/src/gamestate.js
```diff
@@ -210,12 +210,6 @@ class GameState extends EventEmitter {
     return this.#burnCardIds.includes(cardId.toString());
   }
 
-  // mixmix - what is this? it doesn't seem to be used for anything
-  isAutoremovableAutopick(cardId, picksPerPack) {
-    if (this.#pickCardIds.length == picksPerPack && picksPerPack > 1){
-      return this.#pickCardIds[0] == cardId.toString();
-    }
-  }
```
mixmix, line 218: @HerveH44 Can you please check this - I think this was safe to remove, but I didn't fully understand what it was intended to be helping do (nothing that I could see)

HerveH44: I think it comes from another feature (multi autopicks) and this one was just checking if the cardId is the one that will be remove if I select another card. 
I am sure you can delete it. It must be a leftover of a refactor.

tooomm:
Is it possible to move the "picked" and "burned" marker to the left edge instead the top to not cover the card name once you selected cards?

mixmix:
Possible yes, but the bookmark icon doesn't make as much sense on the side, unless it's rotated?

I chose that location because of anything the name is the least functional part of a card (if you're reviewing your selection) and someone mentioned i should think about the text view when designing this (putting it on the left would obscure card text there).

Other ideas would be to change the border color only, but without the symbol that's not very accessibility friendly (color blindness, and also people learning the ui)

tooomm:
The icon should get adjusted a bit to look best at a different location (e.g. rotate), yes.

I consider the name of a card the most important part in order to identify it. That's why I prefer to cover fractions of the art instead rendering the name mostly unreadable.
If you discuss picks or builds you also use names over describing what a card does in most cases. :)

Text-only, hmm... I did not check how it looks there.
Those bookmarks could be hidden on card hover. But The current pick/burn select options on hover are already covering text.

Instead the top left (or left top as I suggested), it could be shown at the bottom left as well. There should be less issues in both modes?

HerveH44:
Seems really good, well done @mixmix . Though, I am the only one not seing the "Card Helper" at the bottom right on the screen anymore ?

mixmix:
The card helper exists for the cards in your deck. Did it ever exist for the ones in the pack?

If it didn't, we should do that in a different pr. If it did, I can fix

mixmix:
@tooomm here's what it looks like with text

image

We basically need to decide what we want to prioritize. Personally I identify cards more by art at a glance, and if I am re-checking a card I want to be reading the rules text. Neither of us is wrong, we just think differently. I am happy to change but want to be super clear whether we care to support text or let it be a secondary citizen.

Another consideration with moving that tag is "where are the pick/ burn buttons when a card is already selected?" we don't want to make it a visual mess with these things overlapping.

**proposal** - try for a design which puts the pick/burn mark in the **bottom-middle**. This is a spot that has nothing in it generally, and less likely to collide with any text. It's a spot sometimes used to put a holo sticker on rare cards...

mixmix:
I might use this up arrow as a colored background
image


https://ionicons.com/v2/ is where the icons we are choosing come from

frontend/src/gamestate.js
HerveH44: How sorry, "I" am going to delete it. Didn't see the comment was on the main PR xD ! Pushing the delete ASAP

HerveH44: Though it's still used by the CardDefault component. I think it should stay there unless you also want to change the behaviour of that Component @mixmix ?

HerveH44:
I don't know why Github doesn't show the deployed app but you can test it here : https://beta-dr4ft-glimpse-draf-wnjww9.herokuapp.com/

frontend/src/gamestate.js
mixmix: It's only in CardDefault because I refractored it there as I was doing this work and was wanting to preserve existing code.

The reason I think it can go is that it is displaying something (I still don't quite understand what), but it wasn't be used by any logic anywhere else to **do** anything.

So is this fine to remove?

mixmix:
Screenshot_2021-01-15 Screenshot
Screenshot_2021-01-15 Screenshot


ok, what do we reckon about this?
I'm feeling pretty happy about it @tooomm @HerveH44

mixmix:
fixed foils (they weren't showing), and took opportunity to do some blending (instead of just overlaying an opaque image):

Non-foil:
Screenshot_2021-01-15 Screenshot
Foil:
Screenshot_2021-01-16 Screenshot

ZeldaZach:
This is looking good! We ready to review/merge? :)

mixmix:
Yes!
Do we have a review process? Probably a light review would be good, as this is going into glimpse-draft and then that (increasingly large) branch should get a good review.


That or we merge things and fix as required

ZeldaZach (approved):
Looks solid to me!

PR #1392, opened 2021-01-18
Fix bug decadent draft not spawning new packs

### LINKED ISSUE #1391: Bug: Decadent draft doesn't spawn new packs.
http://dr4ft.info/
Chrome Version 87.0.4280.141 (Official Build) (64-bit)

----
Launched a new decadent draft.  After each pick, the same pack is presented, with one less card.

HerveH44:
## Linked tickets
- Fixes #1391 

## Explanation of the issue
Since the work on Glimpse draft and the introduction of "burn" concept, we did not take care of burning cards from decadent draft.
In fact, as we chose to consider the decadent draft as X pick, burn the rest of cards, we forgot to burn them really when the pick selection was confirmed (because the player do not have to burn them manually, it's done by the backend).

Move the burn process at confirmSelection and not at handleTimeout because decadent draft do not select burn cards so we have to calculate them when the selection is confirmed.

### DIFF
--- a/backend/player/human.js
+++ b/backend/player/human.js
@@ -131,6 +131,10 @@ module.exports = class Human extends Player {
     // Remove burned cards from pack
     remove(pack, (card) => this.selected.burns.includes(card.cardId));
 
+    // burn remaining if needed cards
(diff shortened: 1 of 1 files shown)


ZeldaZach (approved):
Works, as expected!

PR #1406, opened 2021-01-20
Cobra versions

nex3:
Download card versions from CubeCobra

Rather than using the default card versions for cubes downloaded from CubeCobra, this fetches the versions that the user has specified for their cube. In combination with #1302, this means that dr4ft.info will use the same card images as CubeCobra.

### DIFF
--- a/frontend/src/lobby/CubeList.jsx
+++ b/frontend/src/lobby/CubeList.jsx
@@ -43,16 +43,38 @@ const CubeCobra = () => {
     if (!cubeId) {
       return;
     }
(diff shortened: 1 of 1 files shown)


HerveH44 (approved):

HerveH44:
Super smart, it's really cool!

HerveH44:
@ZeldaZach when you'll merge it, be sure that you reupdated all the sets into the database. The previous PR of @nex3 modified some data, so to be working you need to have these files inside /data up to date !

ZeldaZach:
Will do!

Issue #1016, opened 2020-05-03
Feature Request: Add MTGO + MTG Arena export options


nex3:
When exporting a draft deck to a txt format, split cards such as [Dead/Gone](https://scryfall.com/card/plc/113/dead-gone) are listed with two slashes and spaces between their names, like `Dead // Gone`. Magic Online can't parse this format, and will refuse to import decklists that use it. It wants split card names to have one slash and no spaces between the words, like `Dead/Gone`.

serokellcao:
I think we should keep the default TXT format intact, and instead add more target formats.

Different applications want different formats. For instance, MTGA handles splits completely differently and doesn't know about its own "DLC" sets.

I wrote an exporter for MTGA: http://thoughtflare.memorici.de:22020 (the only exporter I know of that actually works), and I think that we should port the working exporters to JS, increasing the amount of target applications we can interact with.

Having a stable normal TXT format is crucial for that approach.

This is why, I'd say that this is not a bug, but rather a feature request ;)

ZeldaZach:
I would call this a feature request too. All sub-systems (Scryfall, MTGJSON, etc) use the double slash approach. Will rebrand the ticket.

serokellcao:
@nex3, if you can give me a spec for MODO, when I'll be adding MTGA export functionality later this week, I'll implement MODO export too!

nex3:
As far as I know, Magic Online's text format import is exactly the same as the `.txt` format emitted today other than the way split cards are represented.

serokellcao:
@nex3, can you please check how does MODO format adventure cards such as "Murderous Rider"?

Bike-shedding: I'd argue that export UX is anything but low-priority. The fact that I had to make http://thoughtflare.memorici.de:22020 for my play-group's convenience is an evidence. It's something that gets in the way of every single drafter who doesn't use Cockatrice to play the matches.

nex3:
Adventures are formatted as just the creature's name, so `Murderous Rider`.

serokellcao:
Thank you, just as I thought!

Will do my best to implement exports by Saturday!

HerveH44:
Hi @serokellcao how is this issue going? Do you need extra help?

dniemitalo:
The spec for MTGO export is almost the same as the existing text export format. Two changes are needed for the new format:
1) Change the slashes in split cards as outlined above: "fire // ice" becomes "fire/ice" 
[...]

PR #1419, opened 2021-01-21
Fix Cockatrice export and add MTGA/MTGO options

### LINKED ISSUE #1018: Clarify download options
Currently the deck download options in the dropdown are named `cod`, `json`, `mwdeck` and `txt` which is not super intuitive for all users and could be easier and more self explanatory like for example:
`Cockatrice (.cod)`, `JSON (.json)`, `Magic Workstation (.mwdeck)`, `Text file (.txt)`

This was created with https://github.com/dr4fters/dr4ft/issues/1016 in mind.
I don't even know the deck format type for MTGO (`MTG Online (.???)`).

### LINKED ISSUE #1242: Bug: exporting split cards or double faced cards uses the wrong format
**Webpage**: www.dr4ft.info `Running Version: 853ccda (build 2020-09-28) - Using MTGJSON card data v4.7.0 (2020-09-28) and Magic Sealed Data booster rules commit 459802c`

----------------------------------------------------------------------------------

## Detailed Description

When you export any adventure card it'll be exported as just the creature name, while when you export any of the new double faced land cards it'll export with `//` in it. If you then export the decklist to cockatrice it'll show up as unrecognized.

If you then edit your decklist to have the proper cards your deck hash will no longer match, making competitive limited play with these cards impossible.

mixmix:
## Linked tickets
- Fixes https://github.com/dr4fters/dr4ft/issues/1016
- Fixes #1018
- Fixes #1242

Also modified the styling of the export area.
image

Re-reading the original issue I think this UI will need to change again - presuming we need to be able to select a "format" that you are either going to copy or download....

<br>

---

## Screenshots

- original
export-old

- new
export


mixmix:
I on wave an idea for how I want to change the ui

First provide user with choice about where they wanting to use the export :
- cockatrice
- mtgo
- arena
- code (json)

Then present them with relevant options :
- download a file
- copy (if it's text / code output)

The problem with the current ui is it assumes people know what formats they want / need?
We might need to add tooltips for where / how to use some files / content

tooomm:
That looks very promising!

>First provide user with choice about where they wanting to use the export :
>
>    - cockatrice
>    - mtgo
>    - arena
>    - code (json)

I'm wondering why we have the `.json` format. What might people use it for?

I wouldn't drop the simple/basic `.txt` option. Even though its content might be similar to the specific app options.

HerveH44 (approved):

mixmix:
Do not merge this yet!

mixmix:
OK progress :
- have started splitting out different exports into different "formats"
- visually you choose a different export format, then you are presented with different download / copy options
- download and copy output depends on the format and in the "backend" (of the UI) I've split out logic which allows us to customise output more easily

Visually it's looks like this sofar. **input very welcome** - I'm not sure the best way to visually convey the different "formats"

Selection_542
NOTE - MTGO does not have a download method defined, so you can't download, you can only copy

Selection_541
Cockatrice has Download + copy, and the copy will be in a format that cockatrice likes to import from clipboard.
Also note the file-extension is shown for you. With this design you don't have to know what file extension cockatrice etc uses for decks, you just need to know you want to use it in cockatrice

mixmix:
cc @LotusWalker would love feedback on the UI described in previous message

mixmix:
Here's an alternative:

image

NOTE: the logic for ensuring all the exports work right isn't done yet. I intend to write good tests over those

LotusWalker:
Great work @mixmix ! I just tested with a draft and have some feedback:

1) Is it possible to add the Download button to the MTGO export as well? This is quicker than using the clipboard.
2) I used clipboard to copy and paste the draft output, but the split cards are still listed in a format where MTGO won't accept them:

1 Jace, Vryn's Prodigy // Jace, Telepath Unbound
1 Turntimber Symbiosis // Turntimber, Serpentine Wood
1 Kytheon, Hero of Akros // Gideon, Battle-Forged
1 Shatterskull Smashing // Shatterskull, the Hammer Pass

Thanks for taking this on!

mixmix:
Thanks @LotusWalker

I've not done the split format fixing yet. I need to set up tests for each sort of export

Yes I can make a download for mtgo. Can you tell me what file format to make for that?

LotusWalker:
Great, thanks! For MTGO it's just simple TXT format.

tooomm:
> Can you tell me what file format to make for MTGO?

Wizards states you can use txt, dek and csv. I suggest to use `.dek` then as that is probably a file extension uniquely connected with the MTGO app. See [their docs](https://magic.wizards.com/en/articles/archive/importing-and-creating-decks) and https://github.com/dr4fters/dr4ft/issues/1018#issuecomment-623477182.

Is MWS still a thing? Looks like their webpage (http://www.magicworkstation.com/) is not updated in over 3 years. Their news log makes me think they don't even support cards after Ixalan?!

mixmix:
Yeah I had same question about it.
Can probably comment out the code and see if anyone asks for it back.

With Arena I might leave a section there saying "please contribute to this
issue and we can build this"

mixmix:
Hey @LotusWalker can you please attach here an example  .dek file from
mtgo?

LotusWalker:
@mixmix 

Sure, here's a sample .dek file from mtgo attached.

tooomm:
@LotusWalker something went wrong with your upload. It's not included. Can you try again, please?

LotusWalker:
@tooomm I ended up emailing it to @mixmix  directly.

mixmix:
I've used this to export MTGO .dek files I've made for @LotusWalker to try out and give me feedback.

mixmix:
Have tested the export options in cockatice and they seem to be 100% working!

LotusWalker:
Great work @mixmix ! Had some time to test, and here's what I found. 

1) The MTGO-A .dek file you sent could not be imported into MTGO because "Who/What/When/Where/Why is not a valid card name" 

2) The MTGO-B .dek file you sent was imported into MTGO without a problem.

3) I am running into problems with the card Fire/Ice. With the format "Fire//Ice" MTGO does not recognize the card (screenshot1). 
FireIce MTGO

With the format "Fire" MTGO actually inputs only half the card...the "Fire" half, without Ice, even though there is no such card just named "Fire" in mtg. (screenshot2) However, MTGO correctly imports the card if you use the format "Fire/Ice" with one slash. I think this is how MTGO treats split cards.
MTGO - Only Fire!

4) Tested the Fuse card Wear/Tear and the Aftermath card Claim/Fame. They are treated the same as Fire/Ice by MTGO. You need the whole name separated by a single slash, otherwise MTGO just imports half the card! (screenshot3)
Fire, Claim, Wear

Hope this helps, and sorry if this is different from what I initially advised you...was surprised myself when I tried uploading to mtgo using variants of the card names.

mixmix:
Can you clarify - with 3/4 are those with copying and pasting in? (or with a modified dek file?)

LotusWalker:
They were actually from uploading a simple .txt file to mtgo. Let me try testing using a .dek file now.

mixmix:
Is there any "paste from clipboard" option? Or only txt/ dek file import?

LotusWalker:
Ok just checked - MTGO successfully imports with both a .txt file and a .dek file. The split cards just have to be in this format:

1 Wear/Tear
1 Fire/Ice
1 Claim/Fame

There's no "paste from clipboard" option on mtgo...you have to import from a file. This is why we just use the download function on Dr4ft instead of clipboard, because with clipboard its one extra step - paste the deck into a txt file and save it, then upload to mtgo.

mixmix:
Well I'm probably going to disable the copy to clipboard for mtgo then, and only have dek download.

Mtgo will be done then.
I'll make a "raw" txt one which people can use to copy / download and customize to their content

mixmix:
I'll need to jangle the case of not having a catId... Eg that unset card.

LotusWalker:
Disabling the clipboard option for mtgo sounds great.

So, the reason you can't upload the Who/What/When/Where/Why card on mtgo is not because of the format, it's because the UN sets were never released on mtgo. I just tried uploading a deck containing one card from each of the 4 UN sets to mtgo, and none of the cards were recognized (screenshot)

1 Accessories to Murder
1 Duh
1 City of Ass
1 Blacker Lotus

UN Sets not on MTGO

tooomm:
Note: MTGA doesn't allow to load/open files. Import from clipboard is the only way. Just the other way around than MTGO.

mixmix:
## Ready to review!

image

Here's the options available. Under each tab, you get options for Copy / Download which match what each platform can support.
I've done my best to integrate all of the input from multiple threads about how the formats should be. I've also left good tests to make it easy to adjust and extend the rules about how to format output (which will no doubt evolve!).

mixmix:
@HerveH44 are you up for reviewing this? (If not, all good, can you recommend someone who could)

tooomm:
Here are my observations regarding the MTGA export:

- Added basic lands returns this error:
swamp
- Picked basic lands (and snow lands) work just fine. Non-basic lands do also work.
- A deck needs to have at least 40 cards in order to not show a weird error in MTGA and import fails. That means importing a deck without lands (e.g. 23 cards) and planning on adding them in MTGA will not work!
  Ideally, our interface supports the user here in order to have a smooth transition of his deck.

Besides that I tried transform, aftermath and mdfc's - all good and working correctly. 👍 

<br>

Cockatrice / MTGO:
- The deck download is actually a text document (filename.dek.txt and filename.cod.txt)

Did not test further.

<br>

General note:
- The deck hash should regenerate on each new download/copy in case the deck changed
- I would prefer this ordering and naming `Cockatrice | MTGA (or MTG Arena) | MTGO | Text`

mixmix:
Thanks @tooomm appreciate the quick smoke test and write up. 
Will fix those things. 

I hope with the editing down to what wasn't working you changed some line breaks, because.

mixmix:
@tooomm **have made changes:** 
:heavy_check_mark: all lands are correctly coded (M20) lands now
:heavy_check_mark: changed ordering + naming (MTG Arena now)

**Could not reproduce:**
:question: cockatrice files downloading as `.cod.txt`
:question: mtgo files downloading as `.dek.txt`

Can you clarify? When I download I get `filename.cod` not `filename.cod.txt`. Inside (if opened with a code / text editor) I see some xml which is a bunch of different rows with detail about cards.

**Things which I don't plan to change** in this PR:
- changing the hash to be only cockatrice export
- alert people when they export a deck that's <40 cards
    - this would already have been broken, I'm feeling like I've done enough on this and want to merge what's done so far

backend/import/toBoosterCard.js
```diff
@@ -48,7 +47,12 @@ const toBoosterCard = (setCode) => (mtgjsonCard, index, rawCards) => {
     manaCost: manaCost || "",
     rarity: upperFirst(rarity),
     url: url || `https://api.scryfall.com/cards/${identifiers.scryfallId}?format=image`,
-    identifiers: identifiersMini,
+    identifiers: {
+      scryfallId: identifiers.scryfallId,
+      mtgoId: identifiers.mtgoId
+      // needed for exporting cards to MTGO .dek
+      // NOTE: not all cards seem to have this identifier..?
```
HerveH44, line 54: @ZeldaZach I think this point is for you. How do MTGJson produce the identifier mtgoId ?

frontend/src/basiclands.js
```diff
@@ -0,0 +1,176 @@
+/* uncomment this and run this file to output lands! */
```
HerveH44, line 1: Should this be kept in the code ? It seems just a momentaneous helper. But this is fine if it's needed.

frontend/src/components/Select.jsx
```diff
@@ -3,21 +3,28 @@ import PropTypes from "prop-types";
...
-    )}
-  </select>
-);
+const Select = props => {
+  const {
+    link,
+    opts,
+    value = App.state[link],
+    onChange = (e) => App.save(link, e.currentTarget.value),
+    printOpt = opt => opt,
```
HerveH44, line 12: I guess this should also go to the propTypes ? It's a way to customize how to render the opt ?

frontend/src/events.js
```diff
@@ -46,18 +48,23 @@ const events = {
     App.update();
   },
   copy() {
+    const {exportDeckFormat: format } = App.state;
     const textField = document.createElement("textarea");
-    textField.value = filetypes.txt();
+    textField.value = exportDeck[format].copy(null, collectDeck());
```
HerveH44, line 53: Is it possible to not pass a `null` here? It always seems awkward to me to pass null in methods, as it's not very clear at first glance what you're supposed to pass.

frontend/src/export/index.js
```diff
@@ -0,0 +1,22 @@
...
+  downloadExtension: ".json",
+  copy: toJSON
+};
+
+export default {
+  cockatrice,
+  mtga,
+  mtgo,
+  text,
+  json: process.env.NODE_ENV === "development" ? json : null
```
HerveH44, line 19: What is the problem with always outputting the json (not only on development mode) ? If it's fine for you, I'd keep it as always outputting the json.

frontend/src/export/magicWorkstation.js
```diff
@@ -0,0 +1,23 @@
...
+//         const [code, name] = cardName.split("|");
+//         const sanitizedName = name.replace(" // ", "/");
+//         arr.push(`${prefix}${count} [${code}] ${sanitizedName}`);
+//       });
+//     });
+//     return arr.join("\n");
+//   },
+//   downloadExtension: '.mwdeck'
+//   // copy ??
+// }
```
HerveH44, line 23: @ZeldaZach would you keep that ? I have no idea if this is kept. My idea would be to delete it.

frontend/src/export/mtga.js
```diff
@@ -0,0 +1,38 @@
+import { ZONE_MAIN, ZONE_SIDEBOARD } from "../zones";
+
+export default {
+  name: "MTG Arena",
+  // download,
+  // downloadExtension: "???"
```
HerveH44, line 6: Could you put a note here to indicate why you commented these lines ?

frontend/src/export/mtga.spec.js
```diff
@@ -0,0 +1,38 @@
...
+import mtga from "./mtga.js";
+
+describe("export/mtga", () => {
+  it("download of *** file", () => {
+    assert.equal(mtga.download, undefined, "download disabled");
+  //   assert.equal(
+  //     mtga.download("My deck", deck),
+  //     downloadOutput(),
+  //     "correct output"
+  //   );
```
HerveH44, line 14: Why is this commented? Should it be deleted ?

frontend/src/export/mtgo.spec.js
```diff
@@ -0,0 +1,43 @@
+const {describe, it} = require("mocha");
+const assert = require("assert").strict;
+
+import { exampleDeck as deck } from "./test-cards.js";
+import mtgo from "./mtgo.js";
+
+describe("export/mtgo", () => {
+  // console.log(multiFacettedCards.map(c => ([c.name, c.layout])))
```
HerveH44, line 8: Maybe deleting these artifacts could be nice ?

frontend/src/export/mtgo.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    );
+  });
+
+  it("copy to clipboard", () => {
+    assert.equal(mtgo.copy, undefined, "copy to clipboard disabled");
+
+    // assert.equal(
+    //   mtgo.copy("My deck", deck),
+    //   copyOutput(),
+    //   "correct output"
```
HerveH44, line 23: Same question, should it be deleted or kept ?

frontend/src/game/DeckSettings.jsx
```diff
@@ -4,19 +4,31 @@ import PropTypes from "prop-types";
...
+
+const DeckSettings = () => {
+  if (App.state.didGameStart || App.state.isGameFinished) {
+    return (
+      <div className='DeckSettings'>
+        <LandsPanel />
+        <ExportDeckPanel />
+
+        {
+          App.state.isGameFinished && /draft/.test(App.state.gametype)
```
HerveH44, line 20: I know in the past we took a bad habit at putting business logic inside components. Do you think this condition should be handled elsewhere, like in the app object? IMO components should be the least aware of business objects, they may just know that the game can be a "endedDraft" but if this is too much to do, it's fine. The majority of components already use this kind of things.

frontend/src/gamestate.js
```diff
@@ -154,7 +135,7 @@ class GameState extends EventEmitter {
     remove(zone, (c) => c.name === card.name);
     // add n land
     range(n).forEach(() => zone.push(card));
-    this.#landDistribution[zoneName][card.colorSign] = n;
+    this.#landDistribution[zoneName][card.colorIdentity[0]] = n;
```
HerveH44, line 138: Are you sure the colorIdentity is always right ? @ZeldaZach could you tell us if there's no caveats with colorIdentity that may not represent the true colors of the card (like devoid or else) ?

frontend/src/init.js
```diff
@@ -4,6 +4,7 @@ import { default as router } from "./router";
 
 App.init(router);
 
+
```
HerveH44, line 7: Why another line here?

frontend/style.css
```diff
@@ -497,7 +497,7 @@ input[type="checkbox"], input[type="radio"] {
 }
 
 .game .fieldset {
-  display: flex;
+  /* display: flex; */
```
HerveH44, line 500: Should it be deleted ?

package.json
```diff
@@ -31,9 +31,8 @@
     "nodemon": "nodemon ./app.js -e html,js --ignore built --ignore frontend",
     "nodemon:debug": "nodemon --inspect-brk=1338 ./app.js -e html,js --ignore built --ignore frontend",
     "lint": "eslint --ignore-path .gitignore .",
-    "pretest": "npm run lint",
```
HerveH44, line 34: What happened to pretest ? We don't use it anywhere?

HerveH44 (approved):

HerveH44:
@mixmix Well done, it's really nice. I gave my feedback but the whole piece is already good to go.

backend/import/toBoosterCard.js
mixmix: I just deleted my data and restarted the server. Send to build and grab the mtgoId fine

mixmix: Oh someone told me that some cats eg unset were never released on mtgo so they have no id there!

frontend/src/basiclands.js
mixmix: I want future devs to know its easy to change the default lands included to a different set - eg m22

frontend/src/export/magicWorkstation.js
mixmix: It depends whether anyone still uses mtg workstation. I know I and my friends don't but that's a small sample.

I figure disable it in the ui and see if anyone raises an issue. If they do we can resurrect this. But this is git. We could just make a clear commit "delete mtg workstation export" and the code is still findable

frontend/src/events.js
mixmix: I have the copy method take both `(filename, deck) ` in anticipation there might be a copy format which has a spot for the deck / file name.

This export format doesn't so I just didn't bother. Could be clearer to pass it in every time though

frontend/src/export/index.js
mixmix: This json is different to the old one - it contains all the info the app is tracking about each card in the deck (set Code, identifiers, cmc, card number,....)

I like it on for debugging, but it might be pretty confusing for non coders.
For a "raw" format i made "text" output, which is simple and very readable.

Up for experimenting and turning it always on if you want

frontend/src/export/mtga.js
mixmix: Yes!
Basically showing that in the future there could be download options of mtga starts storing a file format.

frontend/src/export/mtgo.spec.js
mixmix: Agree. Thanks will do

frontend/src/export/mtgo.spec.js
mixmix: Those is again hinting about what code should be there to help a future person adding a test. It helps it fit the pattern with other exports.

Happy to delete if it's not a useful breadcrumb

frontend/src/game/DeckSettings.jsx
mixmix: Yeah I feel like I copied this pattern from somewhere. 
I agree it feels like the component knows too much. If this that different than exposing redux state? Some component in the tree needs to have this passed in...

frontend/src/gamestate.js
mixmix: This is a piece I coded by pattern matching. Thinking :
- changed the basic land const, and noticed that had colorSign property
- searched the app and didn't find any other dependences on this variable other than this
- saw that generally cards being passed around by app don't have colorSign
- decided to use colorIdentity which is already on the cards im basic lands.js because **assumed those were the only cards we were looking for a color sign on**

Might need someone to check that's true. These #properties look mysterious to me

package.json
mixmix: I didn't like that linting happened before the tests when you ran npm test so I move npm run test **into** the test script. Basically i want to focus on my tests passing not on semi colons first. I want linting throwing errors _after_ everything on tests is passing. 

So now the tests run, and then the linting runs.

Happy to undo if wanted

tooomm:
> ✔️ all lands are correctly coded (M20) lands now

Works!

I realized in testing, that it's possible to simply include working lands without any set info, like this:
```
10 Plains
10 Island
10 Swamp
10 Mountain
10 Forest
```

MTGA then uses a generic "Arena" set icon:
arena basics

Maybe that's an option to make things a bit simpler?

<br>

> **Could not reproduce:**
> ❓ cockatrice files downloading as `.cod.txt`
> ❓ mtgo files downloading as `.dek.txt`
> 
> Can you clarify? When I download I get `filename.cod` not `filename.cod.txt`.

The root for the confusion was that if I click on download and select `Open with` once Firefox asks me what I want to do with the file, it downlaods to the temp folder and opens it. The file got stored as `.cod.txt`/`.dek.txt` then (note the headline).
If I opt for `Save` it places the correct extension:
confusion


<br>

>Things which I don't plan to change in this PR:
>
>    - changing the hash to be only cockatrice export
>    - alert people when they export a deck that's <40 cards (Note: MTGA only issue as they have this import check!)

Absolutely agree.
I just like to note down related things when reviewing. That way nothing is forgot and new issues can easily be created.


Thanks again for all your work and this nice improvements!!

frontend/src/export/text.js
```diff
@@ -0,0 +1,24 @@
+import {ZONE_MAIN, ZONE_SIDEBOARD} from "../zones";
+
+export default {
+  name: "text",
```
tooomm, line 4: I think it looks nicer when buttons use capital styling --> `Text`

frontend/src/export/mtga.js
tooomm: See the note you put [here](https://github.com/dr4fters/dr4ft/pull/1419/files#diff-7c9a7fdc079f9f1076a8bc832bd135d03f2c65a5ca5ab0c285c47d751af55f52R7):
`  // copy // NOTE - mtgo does not support import from clipboard`

Something along the lines works here, too.

frontend/src/export/magicWorkstation.js
tooomm: I'm also not aware of people still using or mentioning it.

But as said above, even their page is heavily outdated and everything seems unmaintained.
>Is MWS still a thing? Looks like their webpage (http://www.magicworkstation.com) is not updated in over 3 years. Their news log makes me think they don't even support cards after Ixalan?!

I assume people are way more exited about the new (and probably way more relevant) export options than the missing MWS...

The MWS hash should also get removed then. I vote for a separate "Remove MWS" PR.

frontend/src/export/index.js
tooomm: The json file should not be relevant for end users IMO, but rather distracting and confusing.

frontend/src/export/text.js
mixmix: Sure, can change this

mixmix:
**Land sets**:
- for arena, will probably leave it with m20 lands. It's easy for someone to change that later if they want
- the mtgo dek format expects mtgoId for cards, so written isn't wasted if that's needed for lands

**Download extensions**
- sounds like that's working fine
- might be that the "util download" needs to add mime type better. Feels like it can be done later

**Request for summary**
A reasonable request, but I've spent a lot of hours on this already, and I don't have the time /energy to do this atm :(

frontend/src/basiclands.js
mixmix: Have left a more clear comment with instructions here

frontend/src/components/Select.jsx
mixmix: Yeah not used any more. Reverted that change

frontend/src/events.js
mixmix: fixed

frontend/src/export/index.js
mixmix: I'm gonna leave it disabled in production then

frontend/src/export/magicWorkstation.js
mixmix: I'll raise an issue with it

mixmix: https://github.com/dr4fters/dr4ft/issues/1434

frontend/src/export/mtga.spec.js
mixmix: removed

frontend/src/export/mtgo.spec.js
mixmix: removed

frontend/style.css
mixmix: removed!

frontend/src/export/text.js
mixmix: Done

frontend/src/gamestate.js
HerveH44: the # just indicates that the property is private.

package.json
HerveH44: it makes a lot of sense. Happy to delete it

mixmix:
Is this ready to go?

ZeldaZach:
You're failing Node12 tests, so I can't merge this. Can you fix that up please? :D

mixmix:
I suspect that's an intermittently failing test and not anything to do with me?
Will check, but as a start have pushed a dummy commit to see if I can get tests to run again

mixmix:
Tada, passing
:disappointed:

tooomm:
@ZeldaZach You should put a note to `News` that we support MTG Arena and MTGO export when you deploy this! 🎉

package.json
tooomm: > it makes a lot of sense. Happy to delete it

@HerveH44 I did not quite get your comment, can you clarify please?

Do you think we don't need the lint run here at all?
Because this change only moved and combined it with the line below, nothing is deleted. It still runs with every `npm test` as better seen [here](https://github.com/dr4fters/dr4ft/pull/1419/files/e50371518577c49b46cbd7406f3e8f89982f3d16#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R34).

So if there is a linting error, 3 CI builds will fail. The `test 12`, `test 14` and dedicated `eslint` one [which only runs lint](https://github.com/dr4fters/dr4ft/blob/master/.github/workflows/tests.yml#L87).

Not sure about the local dev process, but I assume we can remove it from test because in a PR's the linting is still checked and one can easily run `npm run lint` locally?

PR #1422, opened 2021-01-24
Rearrange cubable cards with split name clashing

### LINKED ISSUE #1420: Bug: Start//Fire erroneously appears in my cube
Chrome Version 87.0.4280.141 (Official Build) (64-bit)

I imported my cube from CubeCobra into Dr4ft, and was surprised to see the Mystery Booster test card Start//Fire in my draft. This card isn't in my cube. I think it might be replacing Fire//Ice (which is in my cube), but not positive about that. In the list of cards imported, it shows "Fire (uma 225)" which I believe is my Fire//Ice, but which is being swapped by Start//Fire.

HerveH44:
## Linked tickets
- Fixes #1420

## Explanation of the issue
When we create the cards that can be cubable, split cards with name's clashing with other split cards create an issue. 
Fire // ice clashes with Start // Fire and in the end the content of the entry 'fire' depends on which card have been processed last.

We merge the cubable cards with care of clashing: the default value of the card name is calculated based on the priority (expansion > funny, etc). For example, Fire will have its default card being "fire // ice" is the old "fire // ice" and not Start // Fire

### DIFF
--- a/backend/data.js
+++ b/backend/data.js
@@ -166,16 +166,42 @@ const writeCubeCards = (allSets, allCards) => {
         ..._.mapValues(groupCardsBySet(cards), keyCardsUuidByNumber)
       }];
     })
-    .flatMap((pair) => {
-      const names = pair[0].split(" // ");
-      if (names.length <= 1) return [pair];
+    /* Create entries for each side of split or double cards.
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Seems to work as expected!

PR #1447, opened 2021-02-09
patch - fix draft log filename

mixmix:
tiny fix - a tidyup from a variable rename in the export refactor/ feature
the problem was it was downloading files as "undefined-draftlog"


tooomm (approved):
This resolves the issue. 👍

PR #1449, opened 2021-02-09
Cleanup column keys after MWS removal

tooomm:
## Linked tickets
- Related to #1442

## Explanation of the issue
With the removal of Magic Workstation, the MWS deck hash column got removed.
That way there was a gap in the key numbering.

Use consecutive numbering


mixmix (approved):
@tooomm I don't think these numbers matter too much - in react, when there are a bunch of identical looking elements next to each other, it needs a "key" that is unique to distinguish the different `td`

That said this makes the code tidier

PR #1451, opened 2021-02-09
Open links in secure new tabs

tooomm:
## Explanation of the issue
Links took users away from dr4ft page.

Open links in new tab instead.
It's suggested to secure them with `rel="noreferrer"` as well.

### DIFF
--- a/backend/game.js
+++ b/backend/game.js
@@ -577,7 +577,7 @@ module.exports = class Game extends Room {
       this.players.forEach(player => {
         if (!player.isBot) {
           player.exit();
-          player.err("Whoops! An error occurred while starting the game. Please try again later. If the problem persists, you can open an issue on the Github repository: <a href='https://github.com/dr4fters/dr4ft/issues'>https://github.com/dr4fters/dr4ft/issues</a>");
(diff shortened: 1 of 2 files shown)


mixmix:
So what's the problem with these that you've found @tooomm - do they not behave as expected?

tooomm:
The ones I changes in this PR work just fine.

Can you please adjust the ones in the footer (https://github.com/dr4fters/dr4ft/blob/tooomm-patch-8/frontend/src/lobby/Version.jsx#L14) properly to improve them as well?
It needs to be done a bit different there.

tooomm:
> So what's the problem with these that you've found @tooomm - do they not behave as expected?

Now reading this again after two years, maybe I should have better clarified this to begin with. @mixmix 😆 

The current links work of course, nothing to "fix" here. This PR  improves how the dr4ft page links to external sites though.
Opening them in new windows instead overriding the current window/tab and taking the user away from the dr4ft page does not sound like the best solution.

I think the same applies for the informational links to the services we depend on under versions in the footer I linked above.

<br>

Trying to cleanup some simple, old stuff still laying around here.

Issue #1432, opened 2021-02-03
Bug: Dr4ft Crashing


ZeldaZach:
Error seen in the logs, that we should investigate:
```
2021-02-02 11:06:22,924-0500 [INFO] [965382d0-6570-11eb-aff1-1f39dcc7d535] Game 965382d0-6570-11eb-aff1-1f39dcc7d535 started.

    Game State
    ----------
    id: 965382d0-6570-11eb-aff1-1f39dcc7d535
    hostId: 1q1lvbf10cx
    title: KLD
    seats: 1
    type: sealed
    sets: 2XM,2XM,2XM,2XM,2XM,2XM
    isPrivate: true
    picksPerPack: 1
    modernOnly: false
    totalChaos: false
    chaosPacksNumber: undefined
    packsInfos: 2XM / 2XM / 2XM / 2XM / 2XM / 2XM
    players: 1 (Rumo)
    bots: 0
    
2021-02-02 11:06:56,411-0500 [ERROR] [GLOBAL] Could not parse XML file: TypeError: pt.split is not a function or its return value is not iterable - TypeError: pt.split is not a function or its return value is not iterable
    at /var/www/dr4ft.info/dr4ft/backend/import/xml/parser.js:79:35
    at Array.forEach (<anonymous>)
    at Object.parse (/var/www/dr4ft.info/dr4ft/backend/import/xml/parser.js:51:14)
    at /var/www/dr4ft.info/dr4ft/backend/api/sets.js:25:30
    at Layer.handle [as handle_request] (/var/www/dr4ft.info/dr4ft/node_modules/express/lib/router/layer.js:95:5)
    at next (/var/www/dr4ft.info/dr4ft/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/var/www/dr4ft.info/dr4ft/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/var/www/dr4ft.info/dr4ft/node_modules/express/lib/router/layer.js:95:5)
    at /var/www/dr4ft.info/dr4ft/node_modules/express/lib/router/index.js:281:22
    at Function.process_params (/var/www/dr4ft.info/dr4ft/node_modules/express/lib/router/index.js:335:12)
2021-02-02 11:06:56,542-0500 [ERROR] [GLOBAL] Could not parse XML file: TypeError: pt.split is not a function or its return value is not iterable - TypeError: pt.split is not a function or its return value is not iterable
[...]

PR #1453, opened 2021-02-10
TEMP: disable JSON upload of custom sets

mixmix:
@ZeldaZach this disables JSON uploads, but allows XML which I think were working ok...

image


tooomm:
Linking the related issue for reference: https://github.com/dr4fters/dr4ft/issues/1432

Issue #1454, opened 2021-02-11
Bug: MTGO crashes when importing .dek file outputs


JhoiraArtificer:
**Webpage**: 
Files were generated by www.dr4ft.info
**Browser**: 
Variety of browsers used by the drafters

----------------------------------------------------------------------------------

## Background Information

MTGO (updated as of 10 February 2021) experiences a fatal error and crashes when importing .dek output of dr4ft.info events

Draft options were:
* Cube draft
* 3 packs with 15 cards from a pool of 540 cards
* Picks per pack: 1

## Detailed Description

Today (10 February 2021, most recent update), I was acting as the draft "secretary" and importing decklists to trade people cube cards. We have always used the .txt output files before without too much trouble, but this time people sent me .dek outputs from the online draft. When I attempt to import the .dek files to MTGO using the "Add deck" icon in the lower left corner of the window in the Collection view, MTGO experiences a fatal error and crashes. I have attached a .txt file of the full error reporting to this bug.

MTGO also errors when I attempt to import the decklist .dek as a binder using the icon in the lower left corner of the window. The only information provided is a pop-up reading "An error was encountered when reading the file". MTGO doesn't crash.

## Further Information

[MTGO dek import as deck error.txt](https://github.com/dr4fters/dr4ft/files/5963141/MTGO.dek.import.as.deck.error.txt)

All the drafters had already closed the draft event, but on the off-chance the URL can provide you useful information, it is https://dr4ft.info/#g/9f9fff90-6c18-11eb-bbd5-1149a9be737e .

mixmix:
Thanks for the bug report.
Can you please attach a broken dek file, or open it in a text editor and copy the contents in here?

LotusWalker:
Dr4ft Version: e20371a

Saw this bug and decided to run a test, and I had exactly the same issue (a crash) when importing the DEK file into MTGO, as either a deck or a binder.

LotusWalker:
I couldn't attach the broken DEK file from my draft because Github doesn't seem to support it, but here's the contents:

<?xml version="1.0" encoding="UTF-8"?>
<Deck xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <NetDeckID>0</NetDeckID>
  <PreconstructedDeckID>0</PreconstructedDeckID>

[...]

PR #1457, opened 2021-02-15
Export again!

mixmix:
fixes https://github.com/dr4fters/dr4ft/issues/1454:
- changed mtgo export to just use `.txt` files

fixes https://github.com/dr4fters/dr4ft/issues/1242:
- added tests around "meld" layout cards

### DIFF
--- a/backend/import/toBoosterCard.js
+++ b/backend/import/toBoosterCard.js
@@ -48,10 +48,7 @@ const toBoosterCard = (setCode) => (mtgjsonCard, index, rawCards) => {
     rarity: upperFirst(rarity),
     url: url || `https://api.scryfall.com/cards/${identifiers.scryfallId}?format=image`,
     identifiers: {
-      scryfallId: identifiers.scryfallId,
-      mtgoId: identifiers.mtgoId
-      // needed for exporting cards to MTGO .dek
-      // NOTE: not all cards seem to have this identifier..?
+      scryfallId: identifiers.scryfallId
     },
     layout,
     isDoubleFaced,
--- a/frontend/src/export/cockatrice.js
+++ b/frontend/src/export/cockatrice.js
@@ -51,8 +51,9 @@ function correctName (card) {
 
   case "flip":
   case "transform":
+  case "meld":
   case "modal_dfc":
-    return card.name.replace(/\s\/\/.*$/, "");
+    return card.faceName || card.name.replace(/\s\/\/.*$/, "");
 
   default:
     return card.name;
--- a/frontend/src/export/cockatrice.js
+++ b/frontend/src/export/cockatrice.js
@@ -53,7 +53,7 @@ function correctName (card) {
   case "transform":
   case "meld":
   case "modal_dfc":
-    return card.faceName || card.name.replace(/\s\/\/.*$/, "");
+    return card.name.replace(/\s\/\/.*$/, "");
 
   default:
     return card.name;
--- a/frontend/src/export/cockatrice.spec.js
+++ b/frontend/src/export/cockatrice.spec.js
@@ -32,6 +32,7 @@ const downloadOutput = () => (
     <card number="1" name="Murderous Rider // Swift End"/>
     <card number="1" name="Refuse // Cooperate"/>
     <card number="1" name="Riverglide Pathway"/>
+    <card number="1" name="Graf Rats"/>
     <card number="1" name="Dead // Gone"/>
   </zone>
   <zone name="side">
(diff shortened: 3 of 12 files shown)


tooomm:
@LotusWalker Can you please test this and let us know if it works now?

Check it here: https://beta-dr4ft-export-again-flwgcb.herokuapp.com/

tooomm:
With commenting out the other mtgo .dek stuff, do you want to keep this?
https://github.com/dr4fters/dr4ft/blob/master/backend/import/toBoosterCard.js#L53

mixmix:
Good point @tooomm yes I think this change sound be reverted. 

Do you think I should comment it out so people can see where you can get mtgoId, or straight up delete it?

tooomm:
Go for whatever you feel preserves the best overview.

Extending the main .dek comment in `mtgo.js` is also an option?

LotusWalker:
@mixmix Ran two drafts to test this and have some feedback. Thanks again for your great work on this!

1) Several cards didn't import into mtgo because their back side was on the TXT file, not their front side. These cards below did not import because of this:

Archangel Avacyn was listed as Avacyn, the Purifier
Jace, Vryn's Prodigy was listed as Jace, Telepath Unbound
Kytheon, Hero of Akros was listed as Gideon, Battle-Forged
Shatterskull Smashing was listed as Shatterskull, the Hammer Pass
Turntimber Symbiosis was listed as Turntimber, Serpentine Wood

So, transform cards and MDFC cards are impacted by this I believe.

2) For the MTGO download, can you make it just .txt instead of .mtgo.txt? We like to rename the filenames before downloading them to streamline things.

Thanks!

mixmix:
Yeah can change that extension for sure.
I changed the logic to use a field called faceName which might not be as reliable as manually mutating the name of cards.
Will have a go

Thank you again @LotusWalker for your specific feedback. It's really awesome having specific examples cards to add to the tests. Each time we add more cases it gets better.

Honestly I'm quite surprised how fiddly this has been. I specifically remember saying "oh, that will be quite easy"!

LotusWalker:
No worries! Let me know if you want me to test again after you've worked on it some more. Seems like a pain in the ass to get this feature perfect.

mixmix:
wow super annoying!
I've not been able to reproduce this bug in development env,
but have been able to see it deployed (https://beta-dr4ft-export-again-flwgcb.herokuapp.com/)

It seems intermitent? When I enter cards for a cube it seems to be fine, but if I run a draft from a set I see problems......I've made made it so that if you know where to hover we can access the raw json data the output is being built from, so hopefully that will help reveal the problem

mixmix:
OK, got something - turns out `card.faceName` IS NOT RELIABLE. i.e. it's not "the name of the front of the card
Have removed reliance on it in the code.

@ZeldaZach can you feed that back to the mtgJSON team please?

specific example:
```js
    {
        "name": "Vastwood Fortification // Vastwood Thicket",
        "faceName": "Vastwood Thicket"    // <<< expected this to be "Vastwood Fortification" as that's the "front"
    }
```

mixmix:
@LotusWalker see if this has fixed it some could you?

LotusWalker:
@mixmix Works perfectly now, nice!

mixmix:
Given this has been "smoke tested" in staffing, I'm going to go ahead and merge this.

We'll need to keep an eye on when @ZeldaZach deploys next before being able to using in production

tooomm:
>    For the MTGO download, can you make it just .txt instead of .mtgo.txt? We like to rename the filenames before downloading them to streamline things.

@LotusWalker What was the exact issue there? You could still rename them, just that the file extension yielded some (important) additional information. You could still remove that later if you prefer? I'm not so sure that everybody has a workflow like your group.

The problem now is that users can download two types of `.txt` files and an ordinary user can not tell which is designed for which use case once downloaded, which will lead to issues.

I can easily imagine a conversation like this:
A: My import is not working.
B: Did you use the .txt file?
A: Sure. (without noticing that he selected the wrong one)

Of course users should download the right one in the first place, but that will happen.

mixmix:
I'm off same opinion as you @tooomm however, I'm also happy to wait to see how it goes for people. We can change it if we get confused feedback?

LotusWalker:
@tooomm Yeah I see your point, but editing the file extensions would have just been one more annoying step for our group, and we all love it rn the way it is. Hopefully it’s pretty clear to folks which tab to use when downloading their decks, depending on what software they use to play matches.

tooomm:
I've still a hard time following your argument in the first place @LotusWalker. 🙈 
Why would you even need (or want) to change the "extension" as you noted?
I did not think that useful (yet removable) tag could bother people.
Or is there some functional issue as all your group seemed to change it? That would be a different story.

Issue #1436, opened 2021-02-05
Bug: Random sets still broken?


tooomm:
## Background Information

Example game ID I copied before crashing on dr4ft.info: 42caaa50-6807-11eb-8cfc-67644fe12bbe

Running Version: 8793772 (build 2021-02-03) - Using MTGJSON card data v5.1.0 (2021-02-01) and Magic Sealed Data booster rules commit 0c510a9

## Detailed Description

I tried to open several sealed games with "Random Set" selected for all 6 packs.
In around 2/3 of the games I get the `Whoops! An error occurred while starting the game.` when I hit the `Start` button.

tooomm:
To add to this, I happen to rerun our Action jobs on on a regular basis lately as they seem to randomly fail just to pass afterwards.

In the logs, I spotted an error in tests about RNG set today, I did not copy it though... the rerun passed all tests again. Will watch out for more of those and paste here.

mixmix:
I've seen the unit tests complain occasionally about RNG not existing

mixmix:
```
   1) Acceptance tests for Pool class
       can make all playable sets
         should return a normal booster:
     Error: RNG does not exist
      at map (backend/boosterGenerator.js:9:11)
      at Array.map (<anonymous>)
      at Object.DraftNormal (backend/pool.js:49:6)
      at forEach (backend/pool.spec.js:95:28)
      at Array.forEach (<anonymous>)
      at forEach (backend/pool.spec.js:91:14)
      at Array.forEach (<anonymous>)
      at Context.<anonymous> (backend/pool.spec.js:90:35)
      at processImmediate (internal/timers.js:461:21)


```

https://github.com/dr4fters/dr4ft/pull/1457/checks?check_run_id=1906081656

tooomm:
Do you think this is related to the issue?

mixmix:
No idea, but you mentioned an RNG bug, so I copied what I saw.

mixmix:
GOT IT
or I got something anyway

PR #1461, opened 2021-02-16
bugfix - random sets

mixmix:
fixes: https://github.com/dr4fters/dr4ft/issues/1436

Pretty sure I got the bug :bug: 

**desicription** : found that while generating a "random set", that "random set" was one of the options that could be randomly selected... which causes later stuff to explode when it tries to build a booster from a set that is not a set, but a placeholder for a set

**fix** : make sure "random" is excluded from list of set types which can be selected from

NOTE: this fix also means that "custom" and "funny" types will be excluded from random selection (where it looks like they were not in the past, though the code was try to...)

### DIFF
--- a/backend/data.js
+++ b/backend/data.js
@@ -251,12 +251,11 @@ const getPlayableSets = () => {
   return playableSets;
 };
 
+const SET_TYPES_EXCLUDED_FROM_RANDOM_SET = new Set(["custom", "random", "funny"]);
 const getRandomSet = () => {
   const allSets = getPlayableSets();
-  const allTypes = Object.keys(allSets);
-  delete allTypes["custom"];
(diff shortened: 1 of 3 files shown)


tooomm (approved):
Ha, nice!!

Created like 10 random pools and no issue so far! Also no more funny cards. 👍

mixmix:
Fuck funny cards aye

tooomm:
`Draft Innovation` sets are still part of random?
Those sets should probably be excluded as well. Not designed for normal, especially in combination with other sets.

Not sure about `Starter`.

mixmix:
Can you open a new issue.
There's a function getPlayableSets which has defined what's payable and near the it also talks about types and excludes some
Could crew an easy mod in there

PR #1463, opened 2021-02-16
Remove lint from npm test

tooomm:
## Explanation of the issue

If there is a linting error, 3 CI builds will currently fail.
The `test 12`, `test 14` and our dedicated `eslint` one [which only runs lint](https://github.com/dr4fters/dr4ft/blob/master/.github/workflows/tests.yml#L87). As the test builds might otherwise succeed, one need to manually double check them to be sure what's the issue every time.

I guess we can safely remove it from the tests to improve their expressiveness. Once a PR is created, the linting is still checked super quickly and shown separately... and one can easily run `npm run lint` locally still.

- Remove lint from npm test

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "nodemon": "nodemon ./app.js -e html,js --ignore built --ignore frontend",
     "nodemon:debug": "nodemon --inspect-brk=1338 ./app.js -e html,js --ignore built --ignore frontend",
     "lint": "eslint --ignore-path .gitignore .",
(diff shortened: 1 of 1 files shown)


mixmix:
Great observation!


I'm gong to go ahead and merge this, but also wonder if we should consider using husky - 
https://typicode.github.io/husky/#/ - which makes it easy to add pre-commit hooks. Basically run linting before each commit.

What do you reckon @tooomm?

tooomm:
Maybe something to consider? I did not use it yet... we can always track it in an issue for now.

PR #1466, opened 2021-02-20
Prettify Cockatrice export

tooomm:
## Linked tickets
- Related #1462

When importing decks to Cockatrice via the clipboard, you only see a long single list with no visual breaks.
Now, that there is an empty line between the actual deck and sideboard it's easier navigate in the list and review the import before confirming.

I tested this with some lists and Cockatrice does not care about empty lines.

Fun fact: I realized this way that cards in the `Junk` zone are not exported at all! 🙈 👍


mixmix:
Cool. If you tested it with Cockatrice and it's fine let's go!

PR #1469, opened 2021-02-22
Connect to mtgpoolservice

### LINKED ISSUE #1423: Feature Request: Separate DB from the app
## Detailed Description
Since last year I tok some time to think how we could cut the application (backend) in 2 separate pieces - one being the database with cards, sets ... and the other being the application, or game engine, handling packs and players. 

I've done some POC during last summer. 
Here is the pool service : https://github.com/HerveH44/mtgpoolservice
I also did the game engine here : https://github.com/HerveH44/mtggameengine

I'm more confident in the pool service being a good start rather than also migrating the game engine to mine. It's a bit dated (september) so it needs a bit of work to also support glimpse draft I think.  

Why Go ?
I did them in Go in order to learn the language and discover its features. I think it's main advantage is that it's fast and has low memory footprint compared to Javascript Node. It has also coroutines which in itself is marvelous. 

Happy to discuss this possibility with you!

HerveH44:
## Linked tickets
- Fixes #1423 

## Explanation of the issue
First step of splitting the project in 2, having another service handle the creation of boosters while dr4ft would focus on handling the game business.

I modified the pool so that it primarly tries to connect to the pool service (which is actually hosted in AWS for test, maybe later it would be best to have the services in a docker-compose or something similar) then fallbacks to the current behaviour in case of error.
This approach helps us understand if we can have enough confidence in this new architecture while not damagin user experience.

### DIFF
--- a/backend/game.js
+++ b/backend/game.js
@@ -103,6 +103,10 @@ module.exports = class Game extends Room {
     return this.didGameStart && !this.isGameFinished;
   }
 
+  get isSealed() {
+    return /sealed/.test(this.type);
+  }
+
   // The number of total games. This includes ones that have been long since
   // abandoned but not yet garbage-collected by the `renew` mechanism.
   static numGames() {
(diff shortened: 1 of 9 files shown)


mixmix:
Hey @HerveH44 I raised a question on this issue you didn't reply to. 

Specifically I'm personally worried about not being able to run my own instance easily, nor being able to support maintenance (if you write go) .

I think having a fallback legacy pool generator code is dangerous because it will make 2 sets of logic to maintain, which will diverge if not careful. We could split out a js module and a go module and share a set of test vectors if you want to do that?

mixmix:
# 🌸

_PS I love your drive to simplify state management. It's really good and I'm keen to support._

HerveH44:
yeah you're completely right, it doesn't make sense actually to separate the 2 services if nobody can take over. Well, I'll close this one until somebody feels he can use this idea. 

I don't feel like working on the game state refactor before some time. You can take over if you want otherwise I'll let it there or close it as you feel it's better.

mixmix:
What do you reckon about splitting out a module which is the booster. So still js, but can be run locally on same server or remotely?

I hope I wasn't too much of a downer. I really didn't want to squash your initiative

PR #1470, opened 2021-02-24
Fix custom card images not displaying + allow text in column view card preview

### LINKED ISSUE #1335: Bug: Custom Card Images No Longer Displaying
**Webpage**: www.dr4ft.info

This seems to happen regardless of what draft options are selected, across all tests I've done.

## Detailed Description

Last weekend and previous to that, my custom card images would display in a draft. Now they just display as a white space and create a new card-sized row below them in the draft with their rules text. This is the line in my xml where I have the imgur link, which hasn't changed. ~~The < brackets are [ so they display here:~~
`<set rarity="rare" uuid="" num="" muid="" picurl="(working imgur link here.png)">CUBE</set>`

## Further Information

This console error is coming up in Chrome:
image

Noojuno:
**Fixes #1335:**
	- Adds a check to make sure the scryfall id exists before checking to see if it's an empty string.

**Hide missing image icon:**
	- If a card image has an error loading it will be marked as errored and not draw the image so the missing image icon and alt text are hidden.
hEyQ6RE

**Allow text in column view preview:**
	- Moves the `ImageHelper` from the column view to use `CardBase` instead of just drawing the images. This allows for text cards to have a preview and also for any card display changes to stay consistent across the site.
O8vcREf

### DIFF
--- a/frontend/src/game/Cols.scss
+++ b/frontend/src/game/Cols.scss
@@ -21,7 +21,7 @@
     }
   }
 
-  .helper {
+  .CornerCardPreview {
     position: fixed;
     bottom: 0;
     z-index: 101;
@@ -38,9 +38,11 @@
     }
 
     .card {
-      width: var(--card-width);
-      height: var(--card-height);
+      display: inline-flex;
+      position: relative;
+      margin: 0;
       cursor: pointer;
+      padding: 1px;
     }
   }
 }
--- a/frontend/src/game/Cols.scss
+++ b/frontend/src/game/Cols.scss
@@ -20,4 +20,27 @@
       }
     }
   }
+
+  .helper {
+    position: fixed;
+    bottom: 0;
+    z-index: 101;
+
+    --card-width: 240px;
+    --card-height: 340px;
+
+    &.left {
+      left: 0;
+    }
+
+    &.right {
+      right: 0;
+    }
+
+    .card {
+      width: var(--card-width);
+      height: var(--card-height);
+      cursor: pointer;
+    }
+  }
 }
--- a/frontend/src/game/card/CardBase.jsx
+++ b/frontend/src/game/card/CardBase.jsx
@@ -14,7 +14,10 @@ export default class CardBase extends Component {
       mouseEntered: false,
       isFlipped: false,
       url: getCardSrc(this.props.card),
+      imageErrored: false
     };
+
+    this.onImageError = this.onImageError.bind(this);
 
     if (this.props.card.isDoubleFaced) {
       this.onMouseEnter = this.onMouseEnter.bind(this);
@@ -29,7 +32,7 @@ export default class CardBase extends Component {
       url: getCardSrc({
         ...this.props.card,
         isBack: this.props.card.flippedIsBack,
-        number: this.props.card.flippedNumber,
+        number: this.props.card.flippedNumber || this.props.card.number,
       })
     });
   }
@@ -42,6 +45,10 @@ export default class CardBase extends Component {
     });
   }
 
+  onImageError () {
+    this.setState({ imageErrored: true });
+  }
+
   render () {
     const { card } = this.props;
     // at the moment for Text view, you can't see both sides of a card on hover
@@ -51,8 +58,8 @@ export default class CardBase extends Component {
       <div className={`CardBase ${card.foil ? "-foil " : ""}`}>
         <CardBaseText {...card}/>
         {
-          App.state.cardSize !== "text" &&
-            <CardBaseImage mouseEntered={this.state.mouseEntered} imgUrl={this.state.url} {...card}/>
+          App.state.cardSize !== "text" && !this.state.imageErrored &&
+            <CardBaseImage mouseEntered={this.state.mouseEntered} imgUrl={this.state.url} onError={this.onImageError} {...card}/>
         }
 
         {this.props.children}
@@ -67,8 +74,8 @@ export default class CardBase extends Component {
       >
         <CardBaseText {...card} />
         {
-          App.state.cardSize !== "text" &&
-            <CardBaseImage mouseEntered={this.state.mouseEntered} imgUrl={this.state.url} {...card}/>
+          App.state.cardSize !== "text" && !this.state.imageErrored &&
+            <CardBaseImage mouseEntered={this.state.mouseEntered} imgUrl={this.state.url} onError={this.onImageError} {...card}/>
         }
         {this.props.children}
       </div>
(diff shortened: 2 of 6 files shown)


frontend/src/game/Cols.scss
```diff
@@ -20,4 +20,27 @@
       }
     }
   }
+
+  .helper {
```
mixmix, line 24: I'd like this to be called something more specific so the styles never collide.
Let's rename the component something like `CornerCardPreview` or something, AND call the class the same thing (caps and all) . Makes finding it in the code base a breeze

frontend/src/game/Cols.scss
```diff
@@ -20,4 +20,27 @@
...
+
+    &.left {
+      left: 0;
+    }
+
+    &.right {
+      right: 0;
+    }
+
+    .card {
```
mixmix, line 40: We might not need the card container any more if you're using raw CardBase.
I think that component inherits the size from that css var --card-height etc.

I'm mainly keen to get things away from that global style.css wherever possible

frontend/style.css
```diff
@@ -364,13 +364,6 @@ input[type="checkbox"], input[type="radio"] {
 
 /** Zones **/
 
-#img,
-.card {
```
mixmix, line 368: You're still using this class (I think unnecessarily) in your code.

Have your searched the whole code base for uses of this? It was kinda everywhere in the past (though I've been pruning it out!)

frontend/src/game/card/CardBase.jsx
```diff
@@ -80,16 +87,12 @@ CardBase.propTypes = {
   card: PropTypes.object.isRequired,
 };
 
-const CardBaseImage = ({ mouseEntered, url, foil, flippedIsBack, flippedNumber, identifiers, name, setCode = "", number = "" }) => (
+const CardBaseImage = ({ imgUrl, onError, name }) => (
```
mixmix, line 90: 🔥 : I might be missing something (reading on my phone), but it looks like you have removed important functionality here:
- double sided cards should flip when you mouse over them
- if an error in loading image occurs, it should try loading an alt source before falling back to text - see `getFallbackSrc`

I think I know why you were making these changes because we don't want the flipping behaviour for that corner preview?

mixmix (changes requested):
This looks really cool @Noojuno, thanks for this PR. I think having text me the ultimate fallback is great, and using the new Card components across the app feels right (I'm stoked, I recently refactored the app to make those)

There's a couple of minor comments / requests and one really important thing which I think would introduce problems (marked with a 🔥 )

Lmk if you have any questions

# 🐬 🍹 ♥️

frontend/src/game/card/CardBase.jsx
Noojuno: The cards do still flip when you hover over them, it's just handled by `CardBase` rather than `CardBaseImage`. The mouse hover functionality was implemented in both places so I went with keeping it in `CardBase` (which also let me use state for marking the error loading). I may totally be wrong about this though so feel free to correct me!

I did remove the fallback source for images though. The initial removal was because as far as my (somewhat limited) testing showed when I was first looking, it would only set the image to the fallback image temporarily since it was just setting the src on the element not the component and then whenever the component was re-rendered it would cause the whole process to start again. I'll push an update through that reimplements this though.

mixmix: Thanks! 
Sorry I didn't get an update about this comment, just reading this after revisiting the pr.

Oh good idea handling the flip at a higher level. Can then likely use it for flipping the text card a well (hmm i wonder how 2 sided cards are handled in text mode atm) 


Please @ mention me when you'd like another review / have a question @Noojuno

IAmUrza:
Will this fix use the image links provided in the .xml file for custom cards? I tried uploading my custom set yesterday and none of the images were showing.

Noojuno:
Sorry about the delay, finally had a chance to finish this up. There are quite a few changes here so I'll quickly write up the gist of a few of them.

`CardBase` has some pretty significant changes.
- I readded fallback image sources.
- I added a new prop `reversed` that allows you to render a card so it has it's flipped side showing by default. 
- I moved the logic for loading an image to a function `getCardImage()`. This takes a `string` to say which card side to get so you can glance through it easily and know what it's doing but this could easily be changed to a `boolean` if you want to avoid magic strings over slightly more readability. Just a personal preference thing.
- I moved the `z-index` that set the text below the image to just be for the body section as having the text always be below the image broke text only cards in column view.
- And finally, I removed the now unnecessary props from the `CardBaseImage` and simplified that down.

The other changes are more limited.
- I renamed the `ImageHelper` to `CornerCardPreview` as requested.
- I removed more global styles from `style.css`
- I changed the key from the index of the card to either the cards `uuid` or `name` as using the index was causing cards to not re-render when added to the column view as the index wasn't unique to the specific card.

@mixmix

Iniquit:
Hi @Noojuno,
Thanks for putting this together! The custom card image fix is really exciting.
I wanted to give you a heads-up about a weird bug I found when using it locally. 

- It seems to occur when you have at least two copies of the same card in the same sort column in your draft deck (cmc sort seems to be the most reliable)
- Drafting another card into that column or moving all the cards in that column to your sideboard by clicking on the latest/lowest card creates "false duplicates" of the repeated card
- Moving all of the cards from your sideboard back to your maindeck the same way creates more false duplicates
- Reloading the page removes all of the duplicates, but you can repeat the process as above
- The glitch can occur multiple times with multiple different double copies of cards in the same draft

duplication_glitch

duplication_glitch

I hope that's helpful for troubleshooting! I saw no console errors and it seems like you can reproduce the phenomenon pretty reliably by doing a M21 draft and intentionally picking lots of double or triple copies of cards.

mixmix:
At a guess I would check for a map where the compendium need a `key` so that the cards can be uniquely identified within the draft.

Thanks for the summary @Noojuno well review this when I am next free

Noojuno:
Thanks @Iniquit! Amusingly the cause for that bug was actually me trying to fix a very similar bug. That should be all working now so if you get a chance to try it on your end again I would really appreciate it.

mixmix:
This is looking really great. I found a couple of quirks, have pulled this into a branch on this repo: https://github.com/dr4fters/dr4ft/pull/1479

So I can modify a couple of things and merge

mixmix:
Great work @Noojuno , I've merged this into `master` and have deployed to my own instance : http://magic.protozoa.nz/

@ZeldaZach will update dr4ft.info sometime

---

p.s. I see from github you're another kiwi @Noojuno ! Where abouts do you live? If wellington, we should play some time

Noojuno:
@mixmix It looks like your changes broke the `reversed` prop (or at least what used to be called reversed) on the card. Here's an example: 
chrome_YRXwlrDYr7
One of the bottom 2 card preview cards should be flipped here

---

I'm up in Auckland. Hilariously I am also actually not really a huge MTG player, despite the fact that this is an mtg draft site lol.

mixmix:
Oh sorry about that break. I was seeing a bug with the double sided cards. Didn't test with the flip ones. Can you see what logic I broke? Sigh

Will have a look at that soon unless you'd like to @Noojuno

I renamed some variables to try and make it more agnostic about whether or was turning cards over or turning them around

PR #1474, opened 2021-02-28
Better set select

### LINKED ISSUE #1131: Feature Request: Use keyrune to display set symbols
## Detailed Description

As seen in recent design drafts, I like a lot how set icons are displayed in the drop down menu next to the set name for the set selection.

The keyrune package (https://www.npmjs.com/package/keyrune) should have the icons we need.

mixmix:
Fixes #1131

Something which annoys me when starting games is that mile-long set selecter.
This adds a better typing search (which does a fuzzy search on setCode + name!)

I also added the set icons to make it even easier to check you're in the right neighbourhood


mixmix:
Looks like this so far : 
image

mixmix:
@HerveH44 can you tell me why that deploy failed? I'm getting access denied. My heroku account is with address whimful at gmail dot com in case adding me is easier

ZeldaZach:
The Heroku instances are a bit funky sometimes, if you're able to deploy it locally with the same script then I'm fine pushing forward

mixmix:
Will deploy on my personal server once this is ready.

Specifically will run the npm scripts : build, start

mixmix:
This is ready to review.
Not sure why auto deploy isn't working. There's a version of it running here : http://dr4ft.protozoa.nz/

tooomm:
I think this is a great improvement.

However, you have to switch back to your mouse and click the desired result now - even if there's only one entry in the list. Before, I would only type and it directly jumps to the entry.

If the first entry in the list would be pre-selected and enter would confirm it (just like a mouse click) that would help a lot.
Using arrow keys to navigate in the results to quickly select the second entry for example would be cool as well.

mixmix:
Yeah noticed this. ā mark against the current deposited is you have to know exactly how to spell innistrad (for example) or remember it was "shadows over innistrad" or something and type "shadows"

This solution offers fuzzy search which is amazing.

I'll look at the module `react-select-search` and see if there are ways to improve the keyboard interaction.
If not, then we have to decide between different UX devils, or find another component which has all the right behavior

tooomm:
>This solution offers fuzzy search which is amazing.

It is! That's why I favour your new set selection. It's a joy to use.

mixmix:
@tooomm I think I've fixed it ! 
http://dr4ft.protozoa.nz/

(honestly I don't quite see why what I changed fixes things but there you have it!)

tooomm:
Thanks for taking another look, it seems to mostly work now. 👍

mixmix:
Ok, smoke test /qa passed.
Anyone able to code review this?

I'll give it a couple of days then merge if I don't hear anything

ZeldaZach (approved):

mixmix:
Yeeee haww!

PR #1477, opened 2021-03-03
XML parser manacost fixes

Iniquit:
## Explanation of the issue
Hi! Previously, when you uploaded a custom Cockatrice-format set xml file to dr4ft, the new set json file that parser.js created in `data/custom/` omitted the `convertedManaCost` key/value pair and didn't convert the `manaCost` value to bracketed mana symbol notation (`1WW` was written to the json instead of  `{1}{W}{W}` like the mtg.json files for official MTG sets use.)

These discrepancies between the json fset files that parser.js was creating and the json the rest of the app expects caused at least two errors during drafts of custom sets:
- sorting custom cards by CMC didn't work
- the automatic "add lands" button from events.js didn't work if there were custom cards in your deck

This PR fixes both issues by:
- labeling cmc correctly as convertedManaCost in set json files generated by parser.js
- using a regex to match mana symbols in manaCost and add brackets to them in set json files generated by parser.js
- updating the tests in parser.spec.js to expect the new format in both cases

Note on the regex: currently it attempts to capture mana symbols (including generic multi-digit mana costs such as `12` and split mana symbols separated with a `/` (such as `{G/W}`) and exclude whitespace characters. One (possibly positive) ramification of the whitespace exclusion is that split cards (such as Fire // Ice) are written to the set json with a manaCost value of their combined halves (`{1}{R}{1}{U}`). The official mtg.json set json files appear to include both card faces but it looks like only the second halves of split cards currently get used in dr4ft for the purposes of determining which lands to suggest, so it seems like smashing the mana symbols of both halves together for custom cards, like this implementation results in, might actually be more accurate since colors from both sides would be counted when determining lands.

I'm excited to contribute to dr4ft, but I'm also still pretty new to writing code for anybody but myself, so please let me know if there're any functional/logistical/stylistic issues with these changes that I should correct to make it a better fit for the project. Thanks!

### DIFF
--- a/backend/import/xml/parser.js
+++ b/backend/import/xml/parser.js
@@ -72,14 +72,14 @@ function parse(content) {
       colors = [],
       layout = "normal",
       loyalty = "",
-      manacost = 0,
+      manacost = "",
       pt = "",
       side = "a",
       type = "" } = root.version === "3" ? c : c.prop;
     const [power, toughness] = pt.split("/");
     const fixedColors = getTrueColors(root.version, color, colors);
     const fixedType = getTrueType(type);
-    const fixedManaCost = addManaCostBrackets(manacost);
+    const fixedManaCost = addManaCostBrackets(String(manacost));
     const set = jsonSets[setCode];
     set.cards.push({
       name: c.name,
@@ -118,10 +118,12 @@ const getTrueType = (type) => (
 );
 
 const manaSymbolRegExp = /(?:(?<!\/)(?:\d+|[^/\s])(?!\/))|\S\/\S|\/\//g;
-const addManaCostBrackets = (manaCostString) => {
-  let manaSymbolArray = manaCostString.match(manaSymbolRegExp);
-  return manaSymbolArray.map((manaSymbol) =>{
-    return manaSymbol === "//" ? ` ${manaSymbol} ` : `{${manaSymbol}}`;}).join("");
+const addManaCostBrackets = (manacost) => {
+  if (manaSymbolRegExp.test(manacost)){
+    let manaSymbols = manacost.match(manaSymbolRegExp);
+    return manaSymbols.map((manaSymbol) =>{
+      return manaSymbol === "//" ? ` ${manaSymbol} ` : `{${manaSymbol}}`;}).join("");
+  } else return manacost;
 };
 
 const getTrueColors = (version, colorv3, colorsv4) => (
--- a/backend/import/xml/parser.js
+++ b/backend/import/xml/parser.js
@@ -119,11 +119,12 @@ const getTrueType = (type) => (
 
 const manaSymbolRegExp = /(?:(?<!\/)(?:\d+|[^/\s])(?!\/))|\S\/\S|\/\//g;
 const addManaCostBrackets = (manacost) => {
-  if (manaSymbolRegExp.test(manacost)){
-    let manaSymbols = manacost.match(manaSymbolRegExp);
-    return manaSymbols.map((manaSymbol) =>{
-      return manaSymbol === "//" ? ` ${manaSymbol} ` : `{${manaSymbol}}`;}).join("");
-  } else return manacost;
+  if (!manaSymbolRegExp.test(manacost)) return manacost;
+
+  return manacost
+    .match(manaSymbolRegExp)
+    .map((manaSymbol) => manaSymbol === "//" ? " // " : `{${manaSymbol}}`)
+    .join("");
 };
 
 const getTrueColors = (version, colorv3, colorsv4) => (
--- a/backend/import/xml/parser.js
+++ b/backend/import/xml/parser.js
@@ -79,12 +79,13 @@ function parse(content) {
     const [power, toughness] = pt.split("/");
     const fixedColors = getTrueColors(root.version, color, colors);
     const fixedType = getTrueType(type);
+    const fixedManaCost = addManaCostBrackets(manacost);
     const set = jsonSets[setCode];
     set.cards.push({
       name: c.name,
       names: getNames(layout, c.name),
-      manaCost: manacost,
-      cmc,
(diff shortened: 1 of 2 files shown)


backend/import/xml/parser.js
```diff
@@ -116,6 +117,12 @@ const getTrueType = (type) => (
   type.split("-")[0].trim()
 );
 
+const addManaCostBrackets = (manacost) => {
+  let manaSymbolRegExp = /(?:(?<!\/)(?:\d+|[^/\s])(?!\/))|(?:\S\/\S)/g;
```
mixmix, line 121: This should be a constant defined once outside the function

backend/import/xml/parser.spec.js
```diff
@@ -116,12 +116,12 @@ describe("Unit tests for XML Cockatrice Parser", () => {
...
           baseSetSize: 1,
           cards: [{
-            cmc: 4,
+            convertedManaCost: 4,
             colors: ["R", "U"],
             isAlternative: false,
             layout: "split",
             loyalty: "",
-            manaCost: "1R // 1U",
+            manaCost: "{1}{R}{1}{U}",
```
mixmix, line 124: This doesn't seem quite right?
The // means split card with two possible costs

mixmix (commented):
Hi, welcome! I'm not familiar with this part of the app, but have left comments on bits that might be worth attention

mixmix:
Also, if you're up for it, there's a possible update to the xml parser (dependobot opened a pr). If you're in the neighborhood of this and are up for seeing if that's a safe upgrade that would be awesome

mixmix:
https://github.com/dr4fters/dr4ft/pull/1441

backend/import/xml/parser.js
Iniquit: Aha, you're right, thanks!

backend/import/xml/parser.spec.js
Iniquit: Keeping the split card indicator is a good idea, particularly since it'll show up in "text-only" mode. I've updated the AddManaCostBrackets function and manaSymbolRegExp to fix this; the function should now catch and preserve the split card indicator, wrapping it in spaces instead of curly brackets when it joins it to the rest of the manaCost string.

Iniquit:
> 
> 
> Also, if you're up for it, there's a possible update to the xml parser (dependobot opened a pr). If you're in the neighborhood of this and are up for seeing if that's a safe upgrade that would be awesome

Sure thing! I merged in the dependobot pr branch locally and was able to install the updated package, build, use the app, and run `npm run test` without issue. Should I merge that branch into this PR?

mixmix:
Yeah I think so. If xml parsing doesn't work we can always roll back.
I presume this is a feature you use so will spot any problems quickly

Iniquit:
Sounds good! I've merged in that branch and another commit that fixes a couple of errors with nonstandard mana costs (blank and generic-only) that I caught while testing.

mixmix:
Let us know when this is ready for final review + testing. If you could provide an example xml file that would be awesome too.

 If GitHub won't let you attach, paste the content into a GitHub gist?

Iniquit:
Sure! I think it should be ready for review and testing. 
[Here's](https://gist.github.com/Iniquit/54e0aac51c7a2bab6b8c881ab9aad3c3) a gist with a custom set xml file I created for testing that includes ~25 deliberately-weird manacosts (snow, hybrid, letter, phyxreian mana, un-set symbols, etc.)

I've been running that file and random set xml files downloaded from [Planesculptors.net](http://www.planesculptors.net) through my local copy to try to catch errors. ["Gunslingers of Trule"](http://www.planesculptors.net/set/gunslingers-of-trule) is one that seemed useful and representative for testing because it was recently created, has lots of cards, and has the "rarity=" tags that dr4ft expects (some of the older sets on there do not.)

mixmix:
Thanks @Iniquit I'll get to testing and reviewing this in the next couple of days

backend/import/xml/parser.js
```diff
@@ -116,6 +117,15 @@ const getTrueType = (type) => (
...
 );
 
+const manaSymbolRegExp = /(?:(?<!\/)(?:\d+|[^/\s])(?!\/))|\S\/\S|\/\//g;
+const addManaCostBrackets = (manacost) => {
+  if (manaSymbolRegExp.test(manacost)){
+    let manaSymbols = manacost.match(manaSymbolRegExp);
+    return manaSymbols.map((manaSymbol) =>{
+      return manaSymbol === "//" ? ` ${manaSymbol} ` : `{${manaSymbol}}`;}).join("");
+  } else return manacost;
+};
```
mixmix, line 127: _style_  (totally my preference, please feel free to ignore)

I would write that function with an "early return to reduce nesting, and use indenting to show the chained functions
```js
const addManaCostBrackets = (manacost) => {
  if (!manaSymbolRegExp.test(manacost)) return manacost

  return manacost
    .match(manaSymbolRegExp)
    .map((manaSymbol) => manaSymbol === "//" ? ' // ' : `{${manaSymbol}}`)
    .join("");
};
```

Iniquit: Oh, that's awesome, thanks! Much clearer to read too!

Iniquit:
Ack, sorry-- I got the branches all snarled up trying to push my local changes up. I think it's fixed now.
66c73e48df80f4db5330e1ce0d841136b116cdbc should have the style change commit.

Iniquit:
Whew! Got it, finally. Git push --force: not the right tool for the job. 0c231b197fa48a421ff995a1e43cef1aa3d676f6 should have the style fix commit.

mixmix:
This got merge-squashed into master.

Should be live and ready for testing @Iniquit (I meant to deploy a test version to try out those files you provided, but someone merged this !).

PR #1482, opened 2021-03-10
Fixes card issue from #1470

Noojuno:
**Fixes small issue in #1470**
Fixes the issue I found in @mixmix's changes to my card change pr where in the column view hover preview cards weren't being flipped correctly by setting the initial value of `isFlipped` to the value of the `showFlipped` prop.

Before: 
chrome_YRXwlrDYr7

After:
5CMq8vW

### DIFF
--- a/frontend/src/game/card/CardBase.jsx
+++ b/frontend/src/game/card/CardBase.jsx
@@ -23,7 +23,7 @@ export default class CardBase extends Component {
 
     this.state = {
       url: this.getCardImage(DEFAULT),
-      isFlipped: false, // this is relative to this.props.showFlipped
+      isFlipped: this.props.showFlipped, // this is relative to this.props.showFlipped
       imageErrored: false
     };
   }


tooomm:
Thanks for the quick fix and sharp spotting!

I realized a small edge case though:
The first card image one sees should be the front, according to the vast majority of reading direction in the world it's the left picture.
It works like that for double faced cards already as shown in https://github.com/dr4fters/dr4ft/pull/1470#issuecomment-795318341. It should behave the same for all cards with more than one face/side.
In your flip card example above it's the other way around. 🙈 

Any ideas here?

mixmix:
Cab you try this with transform cards (eg innistrad?)

That piece of the logic was not a mistake - I ended up changing logic around this area because I found it hard to reason about.

My idea was that state.isFlipped should be talking about whether it was flipped from the default as defined by props.showFlipped

I want to check the logic in the getImage section again

mixmix:
Got it : https://github.com/dr4fters/dr4ft/pull/1484

PR #1487, opened 2021-03-11
Card text improvement

mixmix:
# Before
Screenshot_2021-03-12 Screenshot(1)

---

# After
Screenshot_2021-03-12 Screenshot

### DIFF
--- a/frontend/src/game/card/CardBase.jsx
+++ b/frontend/src/game/card/CardBase.jsx
@@ -130,8 +130,26 @@ const CardBaseText = ({ name, manaCost, type = "", rarity = "", power = "", toug
 
     <div className="body">
       {
-        text &&
-          <div className="text">{text}</div>
+        text && (
+          <div className="text">
+            { 
+              text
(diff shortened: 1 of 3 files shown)


frontend/src/game/card/CardBase.jsx
```diff
@@ -130,8 +130,26 @@ const CardBaseText = ({ name, manaCost, type = "", rarity = "", power = "", toug
...
+              text
+                .split('\n')
+                .map(line => {
+                  const bracketSection = line.match(/\([^\)]+\)/g)
+                  if (!bracketSection) return line
+
+                  const [before, after] = line.split(bracketSection[0])
+                  return [
+                    before,
+                    <span className='bracket'>{bracketSection[0]}</span>,
```
mixmix, line 145: NOTE - this assume there will only be one bracketed helper text section _per line_. I think this is a reasonable assumption to make for 99% of cards and massively increases the readability of card with a lot of helper text

PR #1497, opened 2021-03-16
add placeholder cards while waiting for packs

mixmix:
image

This is a minor tweak which puts placehoder cards in to show the expected number of cards you will receive when the next pack arrives. It stops the UI jumping around so much when a pack arrives and is visually satisfying as the cards "step in" to the shadows that are there already

### DIFF
--- a/backend/game.js
+++ b/backend/game.js
@@ -17,7 +17,15 @@ module.exports = class Game extends Room {
     super({ isPrivate });
     const gameID = uuid.v1();
     Object.assign(this, {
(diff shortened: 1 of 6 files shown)


mixmix:
urgh, the global state in this app makes me feel sad - it's such an untested mess of variables.
Adding to that it revives state from localStorage with no checks as to the types it pulls in which can break your app ):

mixmix:
If no one else reviews this I'll come back and review it (assuming I'll have forgotten what I did in a couple of days!) and merge this

PR #1498, opened 2021-03-16
CI: Add node.js 15 to the mix

tooomm:
## Explanation of the issue
Node 12 entered maintainance LTS.
Node 14 is the new active LTS version.
Therefore we do no longer test against the current release.

See the Long Term Support (LTS) schedule page: https://nodejs.org/en/about/releases/

- Add Node 15
- Use [suggested version style from docs](https://docs.github.com/en/actions/guides/building-and-testing-nodejs#specifying-the-nodejs-version)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,8 +19,9 @@ jobs:
     strategy:
       matrix:
         node_version:
-          - 12
-          - 14
+          - 12.x
+          - 14.x
+          - 15.x
 
     runs-on: ubuntu-latest
 


ZeldaZach:
Node15 is a beta release, and as such we won't support it. Node16 is our next target.

tooomm:
> 
> 
> **Node15 is a beta release**, and as such we won't support it. Node16 is our next target.

Not really a beta...
https://github.com/nodejs/node/blob/master/CHANGELOG.md
https://nodejs.org/en/
https://nodejs.medium.com/node-js-v15-0-0-is-here-deb00750f278

I've to admit though, that Node 15 will be replaced by Node 16 very soon as the linked release schedule in the PR description points out. 16 it is then 👍

mixmix:
Agree. I'm on 14 at the moment. I think we could upgrade to that

PR #1505, opened 2021-03-25
Dockerfile: Use Node LTS-alpine

tooomm:
## Explanation of the issue

> (https://github.com/dr4fters/dr4ft/pull/1503#issuecomment-806360781)
>Bad bot. LTS versions of node only

Here you go, @mixmix.

- Use the current active LTS image of Node (same "alpine" base as before)

We have to see if dependabot knows about the LTS tag or still tries to update to 16 once available. Then, we can still exclude the Dockerfile from dependabot updates.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.19.0-alpine
+FROM node:lts-alpine
 ENV NPM_CONFIG_LOGLEVEL warn
 
 # Install "git"


mixmix (commented):
I'm not qualified to review docker changes. If you have run this code and it works, then could go for it @tooomm

I would hazard it might be safer to explicitly name an LTS version like 14 than have it auto upgrade to something which breaks things?

PR #1506, opened 2021-03-26
CI: Expose node version used in docker container

tooomm:

PR #1541, opened 2021-07-18
Fix can get data directory test for Windows

psettle:
One of the unit tests was assuming a forward slash seperator,
even though the implementation uses path.join(). This meant the
test would fail on Windows platforms because on Windows path.join()
would use a back slash path seperator.

Type: Bug Fix

### DIFF
--- a/backend/data.spec.js
+++ b/backend/data.spec.js
@@ -1,7 +1,7 @@
 const assert = require("assert");
 const path = require("path");
 const data = require("./data");
-const {describe, it} = require("mocha");
+const { describe, it } = require("mocha");
 
 describe("Acceptance tests for Data functions", () => {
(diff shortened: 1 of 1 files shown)


psettle:
I ran into this issue trying to get the project set up. Perhaps this is a configuration issue on my part, I'm not sure.

tooomm:
I cherry picked your change and included it in a test run of our CI on windows as we normally only run it on Ubuntu Linux:
https://github.com/dr4fters/dr4ft/pull/1506/checks?check_run_id=3096950675#step:5:328
It makes the test pass!

(Withour your change, it failes indeed on windows: https://github.com/dr4fters/dr4ft/pull/1506/checks?check_run_id=3096925547#step:5:409)

Issue #1029, opened 2020-05-13
Improve interface for new game modes


tooomm:
In https://github.com/dr4fters/dr4ft/pull/990 we spottet an issue with how game types are layed out currently. The way we present the options and combinations is limited and doesn't allow for further game implementations.

But it looks like we basically found an agreement about adding a new layer for flexibility and all future possibilities to the game selection matrix already:

--> `Game`, `Pool/Cards` and `Pick` type:

>||||||
>|---|---|---|---|---|
>| **Game Type** | Draft | Sealed | | |
>| **Card Pool Type** (=Booster Content) | Regular | Chaos | Cube | ... |
>| **Picking Type** (specific to draft game type) | Regular | Decadent | Glimpse | ... |
>| more? other? | | | | |
>
>More details and explanation: https://github.com/dr4fters/dr4ft/pull/990#issuecomment-623460900

Only @HerveH44 didn't say what he things about it yet. :)

(Linking @keeler)

---

Mentioned concerns so far:

>@ZeldaZach:
>I like tom's idea of a 3rd category, although it might add some excessive button clicks to the program.

Hmm. Can you elaborate further?
The `Regular` variants would always be preselected of course. So no additional clicks for creating a standard draft or sealed game room. In case one wants to play with a different card pool, that also doesn't involve more clicks as in the current implementation.

With new game modes and variations of Game, Pool and Pick type a lot of combinations are possible which is great. Therefore having "pre-configured options" that combine a combination of pool type and pick type and offer it as "Decadent Draft", "Glimpse Draft", "Rochester Draft" and so on would result in many more buttons and a bad interface.

>@serokellcao:
>Re — unnecessary cicks, I'd just have a text field w/ autocomplete as an
alternative input.

I'm not sure if I got you 100% right. How can a text field help with the needed "clicks" (consider the button hits to type)? Having a drop down menu is also not ideal IMO...

tooomm:
Made a mockup for a potential look in the current UI:
![mockup](https://user-images.githubusercontent.com/9874850/82204572-7086f380-9905-11ea-9656-e69836289b7c.png)

tooomm:
I revisited this after a while and updated the mockups with the new modal redesign:
![new-layout](https://user-images.githubusercontent.com/9874850/127327517-68d557a1-e4b5-4f7d-aa05-1124659ed93c.png)

Options could also be placed next to each other instead below to save space and decrease needed scrolling:
[...]

Issue #1540, opened 2021-07-18
Feature Request: 2 Player Cube Fromat


psettle:
## Detailed Description
I'd like to add support for a 2 player cube draft format I play quite regularly.

The format goes like this:
1. Make 4 packs of 30 cards
2. Randomly assign 1/4 and 2/3 to each player (The 1/4 player picks first from the 1st and 4th pack)
3. For each pack, lay the pack face up on the table
4. Players alternate 10 picks each from the pack (first picks according to order from step 2)
4. Discard the remaining 10 cards from each pack

I figure to support this I'd need to do these things:

1. Increase the max burn value to 10 (or perhaps just remove the limit on burn value)
2. Add an 'single pack draft' mode, where only 1 pack is drafted at a time
3. When the single pack draft mode is enabled, first picks will cycle all N players forward on the first N packs, then cycle back on the next N packs, and so on

Ideally I'd also like to provide a way to view others' picks, as it's a part of the format, but I imagine this may be quite difficult and also wouldn't scale to many players well, so I don't plan to attempt this.

## Further Information

I'd like to add the above and submit a pull request for it, I saw a note to submit a ticket to discuss the feature first though!
Please let me know what you think :)

tooomm:
Hi psettle and welcome to the dr4ft repo! :)

Does this play mode has a name? Never heard about something like that.
With the shared booster it reminds a bit of Rochester Draft (#997).

psettle:
We've never named this format, and I haven't heard of anyone else playing it.

It is pretty similar to the Rochester format, the main difference is the pick order is a little different and Rochester doesn't have burn cards.

My understanding is that Rochester adds a reverse to the pick order in the middle of each pack, where the format I proposed doesn't include that.

psettle:
Did anyone have any other feedback or comments on this idea?

tooomm:
>We've never named this format, and I haven't heard of anyone else playing it.


How do you want to implement it with the UI as it's basically a custom and niche format?
As it's not well known and there is not even a specific name to it you need to explain it to users.
Also if we want to support such specific and less known game types, it must be ensured that there are not going to be 10 in 2 years - all on the first level - and the user is just overwhelmed by them.

I mentioned the official Rochester format because it should probably be considered when developing something with various similarities to allow for a later expansion.

<br>

>Did anyone have any other feedback or comments on this idea?

Things move slow around here recently. :)

@ZeldaZach @mixmix

psettle:
[...]

PR #1543, opened 2021-07-29
[WIP][DNS] Begin abstracting pool source

psettle:
## Linked tickets
- Starts on #1029

Pulls the data related to the card pool into a new class. I'm not really sure if I'm on the right track, but I think this is a bit of an improvement?

Also this change enabled bots on decadent drafts, mostly as a side effect of some of the abstractions I introduced around the idea of 'burn cards'. I think this is just an improvement? I can't think of why you'd want to disable bots explicitly on decadent drafts.

My hope here is that this is a base to start separating the cards in the pool from the format being played.

I think these changes can get us going in the right direction in future patches:

- change PoolBuilder.createPool() to PoolBuilder.createPacks(size, number), and move the logic about what size and number of packs out of the /pool module into a new "/format" or "/gamemode" module
- Change PoolBuilder into an abstract interface, and implement it for [cube, regular, chaos] pool types. Then a poolBuilder can be injected into a Game() object as it is created. 

## Testing

I tested each type of draft manually and it seems OK. This change does break the automated tests horrifically though. I'm kind of looking for some feedback on the approach before I sort that out though.

Please let me know what you think!


backend/game.js
```diff
@@ -603,25 +476,23 @@ module.exports = class Game extends Room {
...
   }
 
   getNextPlayer(index) {
-    const {length} = this.players;
+    const { length } = this.players;
     return this.players[(index % length + length) % length];
   }
+
+  // Accessor required for unit test "can make a draft with 4 sets"
+  // (Note: It's weird that the unit tests evaluate private class members instead of
```
psettle, line 493: The test that is validating this struck me as very weird for how I am used to thinking about and designing unit tests.

For a class/project like this I would have either expected to see unit tests that are testing the entire backend as a unit
(i.e. the unit tests would send fake messages that we expect to receive from the front-end, and validate that the response messages meet the expectations)

and/or

I would expect to see each class/module in the backend treated as a unit
(i.e. injecting mocks of the classes these components depend on, then testing the public API of the class/module while setting expectations on the mocks to stress the expected behaviors of each component in isolation)

(That said my unit testing experience is all on much lower level languages (C/C++) and more safety critical applications, so I'm not sure what the conventions are up here in web land.)

I'm not proposed any major overall of the unit tests. I don't see that as a very practical option; but I think the 2nd approach I listed there can work pretty well on this class with the poolBuilder abstraction I've proposed in the push request going forward? Curious for others' thoughts on this.

tooomm:
@ZeldaZach, @mixmix: Please share your thoughts!

backend/pool/pool.js
```diff
@@ -1,6 +1,6 @@
-const {sample, shuffle, random, range, times, constant, pull} = require("lodash");
+const { sample, shuffle, random, range, times, constant, pull } = require("lodash");
```
mixmix, line 1: this file rename is a little confusing to me.
what is pool/pool? should it be `pool/index.js`?

backend/pool/poolBuilder.js
```diff
@@ -0,0 +1,217 @@
+const Pool = require("./pool");
+const { truncate } = require("lodash");
+
+/**
+ * @brief Manages parameters that control the type, grouping and quantity of cards in a pool
+ */
+module.exports = class PoolBuilder {
```
mixmix, line 7: I love this abstraction, it feels a lot tidier

backend/game.js
```diff
@@ -1,16 +1,16 @@
 const crypto = require("crypto");
 const path = require("path");
-const {shuffle, truncate} = require("lodash");
+const { shuffle } = require("lodash");
 const uuid = require("uuid");
 const jsonfile = require("jsonfile");
 const Bot = require("./player/bot");
 const Human = require("./player/human");
-const Pool = require("./pool");
+const PoolBuilder = require("./pool/poolBuilder");
```
mixmix, line 8: my gut feeling is this should require `pool` and that `poolGenerator` or whatever it's called should be required inside pool so you can just go `pool.create` or `pool.generate` and then that calls out to the booster and pool creating functions.

backend/pool/pool.js
psettle: I moved this into /pool because I am hoping to make /pool a component (module? package? not sure on the terminology here) that exposes PoolBuilder as an interface.

I'm happy to go with whatever the convention is for node.js (I don't really have any node.js experience).

Re: pool/pool, please see my notes about where I see this change going in the pull request description.

I'm hoping to see PoolBuilder being split into 3 different implementations of an abstract interface (cube, standard, chaos) and all the logic currently in pool/pool would be split up and included directly in those 3 interface implementations.

backend/game.js
psettle: I think we have a pretty similar idea of the target design here, but that we are being caught up on naming a little.

I think that PoolBuilder would be more appropriately called 'Pool', but I chose PoolBuilder because I didn't want to create confusion and conflicts with the existing usage of 'pool' in this project (where pool is an array of cards, as in Game.pool).

In future patches I see removing the current usage of 'pool' completely:
- the pool/pool component would be integrated into various implementations of PoolBuilder that would be injected into Game externally
- the current use of 'pool' in Game.js (this.pool) could be more appropriately called (this.packs) or some such
  - (Which would signal that this.packs is a concrete representation of the exact cards to be used, instead of the more abstract idea of a Pool as the rules that define the bounds of the possible cards to be used that PoolBuilder is currently capturing.)

At that point I think we could (and should) rename PoolBuilder to Pool in a way that matches your intuition here.

psettle:
Thanks @mixmix for the review!

A lot of this patch is doing setup for other architecture changes related to https://github.com/dr4fters/dr4ft/issues/1029.

I'm not sure if you would prefer I do all the changes I am suggesting here in a single pull request, or if doing it in a few steps is easier to review/discuss, I'm happy to go either way, please let me know :)

mixmix:
Probably small PRs.
Some context is that this repo doesn't get much active development, and there's only a couple of us reviewing things. So easy PRs are most welcome!

backend/pool/pool.js
psettle: So I did a bit more research here and I see that index.js serves as the entry point for passing a folder into require(). I don't think this file makes sense as an entry point to the 'pool' abstraction (indeed it is not included by anything external to this folder anymore).

I can't think of anything useful an index file would do for this folder, I think it would just make the import in game.js more confusing probably? Please let me know what you think :)

ZeldaZach:
Hey there! Any word on if this is still compatible/finished? :)

psettle:
I think this change is okay if you want it, I don't really have a personal interest in continuing work on the initial feature I proposed (https://github.com/dr4fters/dr4ft/issues/1540). The main motivation for me was gathering restrictions related to the pandemic, and they haven't been activity in my area for a while, so we've just been playing in person for a while now. Feel free to close.

Issue #1148, opened 2020-07-13
Feature Request: Setup heroku review apps


tooomm:
## Detailed Description
We should get our own Heroku setup. Currently, we still use the one from Herve.

## Further Information
https://devcenter.heroku.com/articles/github-integration-review-apps

mixmix:
I think this is done yeah @HerveH44 ?

tooomm:
Herve still runs the Heroku app we use in the repo from his account - Zach has no access to it as far as I'm aware.

This ticket got created after Herve announced that he can no longer dedicate time to the project after basically taking over the project/main development for a while. He is still around from time to time and opens a PR, but not very regularly.
We have no easy way to reach out to him as he is not very active here or Discord I think. [Maybe mail?](https://github.com/dr4fters/dr4ft/blob/master/package.json#L18)
Don't even know if he follows the repo or recognizes mentions.

In order to prevent to lose access to the (working) Heroku preview app for whatever reason it sounds reasonable to run it from within the more active team and have better control and admin access.
Recent events and problems with the deployments stressed that. We can not even debug and look into the logs when problems occur. It's difficult to maintain a properly working integration with no access to the setup.
As it's broken since a few days and we have no clue, this ticket is even more up to date.

Probably @ZeldaZach is the right person to take care of this as he runs the main server and is in charge of the GitHub repo as well?
If I recall Herve correclty, he all ran it from the free trier as the deployments are only active on request, get destroyed after PR's are merged etc. I don't think they have a special offer for open source projects...
(Maybe Zach even runs the main dr4ft instance via a Heroku plan already? That would be super easy then...)

If we would have access to the current integration that should also work as mixmix noted in one of the PR's with failing Heroku deployments.

PR #1550, opened 2021-08-14
set a max-width for lobby + game, center the app

mixmix:
The app looks great on a laptop. On my desktop with a large monitor it's a bit painful to use because the cards get really spread out. Here's my proposed solution

### DIFF
--- a/frontend/src/game/Game.jsx
+++ b/frontend/src/game/Game.jsx
@@ -15,6 +15,8 @@ import {vanillaToast} from "vanilla-toast";
 import "vanilla-toast/vanilla-toast.css";
 import {ZONE_JUNK, ZONE_MAIN, ZONE_PACK, ZONE_SIDEBOARD} from "../zones";
 
+import "./Game.scss";
+
 export default class Game extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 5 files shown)


mixmix:
notice I've chosen a max-width which gives you 5 column. this means that you can look at a 15 card booster quite concisely

tooomm:
I can not  easily check this currently. (https://github.com/dr4fters/dr4ft/issues/1148 🙈)

From the pictures you provided, I like the centred placement. Only 5 cards per row seems reasonable as well.

With that many empty space visible on the sides now, the cards look super tight... Did you try a bigger gap between them?

Does this affect deck building as well?

mixmix:
yes it effects deck building. Ideally that would be able to be wider... 

yeah good point about tightness. That's a little harder to effect because I think I wrote this with `grid-gap` ...
we could change that with media-queries I guess

tooomm:
In deck building, for cmc/mv sorting, color or type, you need 7 piles/columns each.
There will be issues with only 5.

7 for both on the other hand doesn't have the nice proposed pack experience.

mixmix:
The deck building card size is smaller. Need to check how many cols there are there

Issue #1531, opened 2021-05-25
Bug: Changing the card image language doesn't trigger a reload


tooomm:
**Webpage**: dr4ft.info
<!-->
**Browser**: Chrome 90

----------------------------------------------------------------------------------

This leads to the image languages not being changeable at all in e.g. Sealed.

mixmix:
There's a hack to this which is to study to text then to desired quality of image. I think this stimulates change

tooomm:
In Firefox at least, swapping to `Text-Only` and back any quality, does only show the already downloaded pictures.
No re-download in selected language.

Actually, it looks like the quality setting doesn't do anything currently. Starting with e.g. STX in high quality does not give me bad pictures when changing it to low quality. Same the other way around.

tooomm:
Update.

>>There's a hack to this which is to (change the `Card image quality`) to text then to desired quality of image.
>
>Actually, it looks like the quality setting doesn't do anything currently.

Just confirmed that Chromium behaves the same as Firefox, it's not a Firefox issue!
I found out that one has to actually reload the page in order to get the chosen settings to apply and display a different picture quality or language.

The same is also true for `Card image language` btw.

tooomm:
Fixed in #1585

Issue #1490, opened 2021-03-13
Bug: image quality hard to change


mixmix:
Changing the image quality (on mobile, and probably desktop) doesn't work.

Eg start on low quality, turn it up to high, nothing happens. Change the sort mode and you immediately see it adjust to correct level.

Changing to text only mode works, but doesn't help image quality change

tooomm:
I can confirm. This is a recent regression. It used to update and refresh the images immediately after the quality setting got changed.

It can still be seen on the [dr4ft beta deployment](https://beta-dr4ft.herokuapp.com) which runs a build from 2021-03-01 due to the recent Heroku issues.

tooomm:
Fixed in #1585

PR #1585, opened 2022-02-24
fix: card-quality settings

mixmix:
There was a bug where changing the card quality didn't immediately update the quality of images loaded (you had to go via text mode then to a different quality).

This fixes that!

### DIFF
--- a/frontend/src/game/GameSettings.jsx
+++ b/frontend/src/game/GameSettings.jsx
@@ -66,18 +66,19 @@ const SortCards = () => (
 );
 
 const sizeDisplay = {
-  "text": "Text-Only",
-  "small": "Low",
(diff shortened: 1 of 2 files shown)


tooomm:
This should address https://github.com/dr4fters/dr4ft/issues/1490 then 👍 

There is an similar issue with the language selector (#1531).
In #1490 there is still an old, working heroku instance linked that is dated at 2021-03-01... both image quality and image language instantly updates on changes.

mixmix:
thanks @tooomm without commit history of where that heroku is at, it's a bit hard to say what changed. Keen to just move forward.
Happy to add the language change to this PR

ZeldaZach (approved):

PR #1586, opened 2022-02-26
replace node-sass for dart-sass/sass

mixmix:
we were seeing problems with the automated docker builds.

identified it might be `node-sass` with node 16 not interacting well.
This PR attempts to see if replacing the deprecated `node-sass` works

What I did:
- followed the link from `node-sass` docs to `dart-sass` which it recommended (this is just the npm module `sass` ?)
- couldn't use latest `sass-loader` without upgrading webpack, so left that
- spun up dev instance and checked scss files are being compiled :heavy_check_mark: 

Now we wait to see what CI says!


mixmix:
hmmm... well looks like docker is working now.
No idea about that beta deploy. Any advice @HerveH44  ?

ZeldaZach:
I'm looking on how we can upgrade to Node14 for the service

PR #1592, opened 2022-03-07
Simpler footer

tooomm:
## Explanation of the issue
Footer included some not really necessary text.
That resulted in forced line breaks with the more narrow layout and on smaller screens.

Only keep relevant hints and links.
Displays all version information in one line again (at least on desktop)


keeler (approved):
LGTM. This should save considerable space in the footer for smaller screens.

PR #1624, opened 2022-06-01
Refactor docker image and CI pipeline

keeler:
# Summary

Refactors the docker image for better layer caching and refactors the github action CI pipeline to run tests and linting in a docker container which takes the CI runtime from ~7m to ~2m. The steps in the CI pipeline are encapsulated in a Makefile so that it's easier to reproduce those steps locally.

# Details

- Refactors `config/version.js` to prefer using a `VERSION_INFO` env var that gets built into the docker image. This is so the `git` command and `.git` folder don't need to be built into the docker image which improves layer caching and saves some space in the image and docker build context.


.dockerignore
```diff
@@ -1,4 +1,6 @@
-data/*
-!data/scores.json
+**/.git
+**/.github
+data
+built
+doc
 node_modules
-frontend/lib
```
keeler, line 4: Removed `frontend/lib` and `data/scores.json` because they don't appear to exist anymore.

Added ignores for `.git` and `.github` because they really shouldn't be built into the docker image anyway. In particular, having `.git` built into the image means that making a commit required rebuilding the image from the `COPY . .` step that existed previously, even if that commit changed no code (e.g. a README change) which is unnecessary.

Dockerfile
```diff
@@ -1,24 +1,40 @@
...
-RUN apk update \
- && apk add alpine-sdk
+ENV NPM_CONFIG_LOGLEVEL warn
+ENV PORT=1337
 
 # Set working dir as /app
 WORKDIR /app
 
 # Add sources to /app
-COPY . .
```
keeler, line 12: This `COPY . .` step makes it so that the entire repo (minus a few gitignored files/folders) gets copied into the image. This forces unnecessary rebuilds of the image even on changes irrelevant to the functioning of the image, such as README changes.

Docker will cache each of these `COPY` instructions, so their ordering reflects how often they may change. Things that will change more often like `frontend` or `backend` are further down, and re-building the docker image will skip the other `COPY` instructions if the files they would copy are unchanged, saving time in the case of image rebuilds.

Dockerfile
```diff
@@ -1,24 +1,40 @@
-FROM node:lts-alpine
-ENV NPM_CONFIG_LOGLEVEL warn
+FROM node:16.15.0-alpine
+
+ARG VERSION_INFO=noVersion
 
-# Install "git"
-RUN apk update \
- && apk add alpine-sdk
```
keeler, line 6: This was installing `git` and a bunch of other stuff into the image, and the only reason I could find was the `git describe --tags` that gets run in `config/version.js` to grab the latest commit hash/tag to display on the lobby page. Installing all that stuff was slow and unnecessary just to show the latest commit/tag on that page. Now the `git describe --tag` call happens outside the build and gets passed in as an ARG, which sets the environment variable `VERSION_INFO` later on in the build, hence the refactor in `config/version.js`. This cuts down on the time needed to build the image and cuts down on the size of the image somewhat as well.

.dockerignore
```diff
@@ -1,4 +1,11 @@
-data/*
-!data/scores.json
+**/.git
```
keeler, line 1: Previously, the `.git` folder was added to the docker image for the sake of building in a version number, namely a commit hash to display in the footer. The other ignores here leave out files that are irrelevant to the docker image build. Any time these files changed required the entire image to be rebuilt.

.github/workflows/tests.yml
```diff
@@ -13,55 +13,28 @@ on:
...
 
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-
-      - name: Setup
-        uses: actions/setup-node@v2.5.1
         with:
-          node-version: ${{matrix.node_version}}
+          fetch-depth: 0
```
keeler, line 25: Setting `fetch-depth: 0` makes it so `git describe --tags` works in the `make docker` step.

Dockerfile
```diff
@@ -1,24 +1,47 @@
 FROM node:lts-alpine
+
+ARG VERSION_INFO=noVersion
+
 ENV NPM_CONFIG_LOGLEVEL warn
+ENV PORT=1337
 
-# Install "git"
```
keeler, line 4: Installing `git` was apparently done only to allow running `git describe --tags` in `config/version.js`. This is slow and bloats the size of the image, and is IMHO unnecessary just to get a commit hash in the footer.

Dockerfile
```diff
@@ -1,24 +1,43 @@
...
+
+ENV NPM_CONFIG_LOGLEVEL warn
+ENV PORT=1337
 
 # Set working dir as /app
 WORKDIR /app
 
-# Add sources to /app
-COPY . .
-RUN adduser -S dr4ftuser
```
keeler, line 13: Node images already include a non-root user called `node`, so adding another one is redundant. https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user

Dockerfile
```diff
@@ -1,24 +1,43 @@
...
+ARG VERSION_INFO=noVersion
+
+ENV NPM_CONFIG_LOGLEVEL warn
+ENV PORT=1337
 
 # Set working dir as /app
 WORKDIR /app
 
-# Add sources to /app
-COPY . .
```
keeler, line 12: Running `COPY . .` is IMHO an anti-pattern. When building a docker image instructions (e.g. `RUN`, `COPY`, etc.) will be cached as a 'layer' in the image. When running `COPY . .` any file changing invalidates the cache for this layer. The restructured so files that likely change more often (e.g. `frontend/` or `backend/`) are part of later layers. This means that if you're only changing the code in `frontend/` or `backend/`, repeated builds of this image will take advantage of the caching of earlier layers and save you re-building them.

Dockerfile
```diff
@@ -1,24 +1,43 @@
...
-RUN adduser -S dr4ftuser
-RUN chown dr4ftuser -R .
```
keeler, line 14: Running `chown` manually like this has a sneaky behavior where it actually copies files in layers above it, increasing the size of the image. The `COPY` instruction includes a `--chown` flag that does not trigger this behavior.

Dockerfile
```diff
@@ -1,24 +1,43 @@
...
-RUN npm ci
+RUN npm ci --ignore-scripts
+
+# Update card database
+COPY --chown=node backend/core backend/core
+COPY --chown=node config/ config/
+COPY --chown=node scripts/ scripts/
+RUN npm run download_allsets \
+ && npm run download_booster_rules \
+ && chown node -R data/
```
keeler, line 23: Including this `chown` in the same `RUN` instruction does not increase the size of the image like it would if it were in its own separate `RUN` instruction.

Dockerfile
```diff
@@ -1,24 +1,43 @@
...
-# Add sources to /app
-COPY . .
-RUN adduser -S dr4ftuser
-RUN chown dr4ftuser -R .
-USER dr4ftuser
+COPY --chown=node package.json .
+COPY --chown=node package-lock.json .
 
 # Install the dependencies
-RUN npm ci
```
keeler, line 18: Running this at the end, in combination with the `COPY . .` meant that any time any source file changed, the dependencies would need to get reinstalled and the card info would be re-downloaded again regardless of whether it was necessary due to the layer cache invalidation at the `COPY . .` layer. This slowed down repeated iterations of docker builds significantly.

Dockerfile
```diff
@@ -1,24 +1,43 @@
...
+
+# Update card database
+COPY --chown=node backend/core backend/core
+COPY --chown=node config/ config/
+COPY --chown=node scripts/ scripts/
+RUN npm run download_allsets \
+ && npm run download_booster_rules \
+ && chown node -R data/
+
+# Add sources to /app
```
keeler, line 25: I tried to order these in such a way that more frequently-changing files would be copied in later layers so the layer cache invalidation would only apply to lower layers. Open to suggestions.

Also, I think the `app.json` file could be excluded here.

tooomm:
That's a bunch of changes again, nice!

Looks like you combined everything in one docker run.
That also means we no longer test against various Node versions to ensure compatibility and spot issues.

>The CI checks now run in ~2m instead of ~7m

I'm looking at improving the CI runs since a while in #1506 and they finish in 1min: https://github.com/dr4fters/dr4ft/actions/runs/2530256044
Total time is ~2min as they run in parallel.
But they do test against all actively maintained and available node versions, including the given one in the .nvmrc file.

What do you think about the node test matrix?
Nothing set in stone here. But maybe its worth combining those?

>Refactors the docker image for better layer caching and (...) the size of the docker image is about 70% what it was before (796MB --> 564MB).

The docker improvements and the reworked layering and caching sounds useful when running the app in a container.
But I can't judge too much about that.

keeler:
> I'm looking at improving the CI runs since a while in #1506 and they finish in 1min: https://github.com/dr4fters/dr4ft/actions/runs/2530256044 Total time is ~2min as they run in parallel. But they do test against all actively maintained and available node versions, including the given one in the .nvmrc file.

> What do you think about the node test matrix? Nothing set in stone here. But maybe its worth combining those?

Nice, I wasn't aware of that PR, thank you for sharing.

I don't understand the need to test against different node versions. Is it because developers might be using an older version of node? Wouldn't establishing and clearly documenting the specific node version to use be much easier than testing and supporting a bunch of different node versions? Ultimately, the app will run using a single node version (current LTS) and that's the one worth testing, and IMHO developers should have no expectation that a different major version of node will necessarily work. Am I being too uncompromising?

Mainly, I replaced the CI steps with docker because it's much easier to reproduce the entire CI pipeline locally when/if there are issues. Also, I assume https://dr4ft.info runs from a docker container? If so, to me that is more reason to build & run everything in the CI pipeline from docker.

> The docker improvements and the reworked layering and caching sounds useful when running the app in a container.

The main benefits of the changes in this PR are:

1. Can repro CI pipeline locally.
2. Faster to iteratively re-build the docker image while developing or addressing CI issues.
3. Smaller image, so faster to push/download and lighter on disk space.



If you feel strongly about testing different node versions, it would be possible to combine our ideas (test multiple node versions + run everything from docker) by using different base images in the docker build, e.g. `docker build --build-arg BASE_IMAGE=${{base-image-from-matrix}}` with a Dockerfile like this:

```Dockerfile
ARG BASE_IMAGE=node:lts-alpine

FROM $BASE_IMAGE
...
```

tooomm:
I don't think Zach runs dr4ft.info from a docker container. Generally everything here was not centred around docker due to lack of knowledge. The Dockerfile is only added to the CI to have at least some basic commands run and see if it works at all (and keeps working).

Isn't it common practice to test against various Node versions in CI?
It's not only about devs, but also people running the app. They might have a different version or other services with different requirements on the same machine. Maybe I'm wrong?
Docker should not be mandatory for sure, but an option.

But that are things devs working with the code and actual web and docker ninjas need to answer.

ZeldaZach:
With a docker container, we only need to theoretically test against the version the Docker container supports. I'd consider switching over to a container system for the main deployment once it's stable. Thanks for the CR, if it's ready for review please let me know!

keeler:
> Isn't it common practice to test against various Node versions in CI?

For libraries/packages intended for use by other projects, definitely yes. For web apps like this it's more common to pick a node version and test only that version (in my experience).

> It's not only about devs, but also people running the app. They might have a different version or other services with different requirements on the same machine.

Doesn't `nvm` solve that problem already? Wouldn't a container image solve that problem as well? Have users asked for support across different node versions? Or is this an assumption?

> Docker should not be mandatory for sure, but an option.

I think we can agree on this.

PR #1701, opened 2023-04-19
Better bots

mixmix:
This PR creates bots that do more than pick randomly:
- for the first 6 picks, it chooses the rarest card in the pack
    - if there are ties it choose randomly among them
- after 6 picks, it choose it's 2 colors - these are just the 2 most common colors from pick so far
- for subsequent picks the bots chooses the rarest cards in it's colors
    - if it can't, then it choose the rarest card (or was it a random card)

This is a super simple improvement. Lots of space for improvement AND play-testing suggests it's already a massive improvement - you don't get bots wheeling bombs and random players getting way overpowered... but you do get bombs wheeling in packs 2-3 more because the bots aren't in those cards colours.

you can try it out here : http://magic.protozoa.nz:1337

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,25 +4,27 @@
 
 version: 2
 updates:
-  # Enable version updates for npm
-  - package-ecosystem: "npm"
-    # Look for `package.json` and `lock` files in the `root` directory
-    directory: "/"
-    # Check the npm registry for updates once a week
-    schedule:
-      interval: "weekly"
-    # Limit the amout of open PR's (default = 5, disabled = 0, security updates are not impacted)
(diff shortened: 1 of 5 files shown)


ZeldaZach:
We should be passing our tests first thought @mixmix

mixmix:
I have a bad feeling there is an intermitently failing test.
Oh yeah its a pack making function - not all packs are unique cards for some sets. I started looking into it, but it was a bit opaque

mixmix:
Ok my bad there were some linting things with semi-colons :vomiting_face: 
But we still have an intermittently failing test on Pool creation for some sets. It's in `master`

mixmix:
arg... no idea why this is failing...

image

PR #1702, opened 2023-04-25
bugfix - fail at getting uuid more gracefully

mixmix:
was seeing CI fail on this particular line in another PR, this is an attempt to fix that

### DIFF
--- a/scripts/download_booster_rules.js
+++ b/scripts/download_booster_rules.js
@@ -37,9 +37,9 @@ async function fetch() {
         acc[code] = {
           balance_colors,
           totalWeight,
-          cards: Object.entries(cards).reduce((acc, [cardCode, weigth]) => {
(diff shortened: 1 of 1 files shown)


scripts/download_booster_rules.js
```diff
@@ -77,8 +77,12 @@ const getCard = (cardCode) => {
 
 const getUuid = (cardCode) => {
   const [setCode, cardNumber] = cardCode.split(":");
-  const { cardsByNumber } = getSet(setCode.toUpperCase());
```
mixmix, line 80: this line here was the gotcha - destructuring `undefined` was causing an error to be thrown in CI

I don't know why the set would not be found, maybe that's an error which should throw, as it shows your data is somehow really wrong.

mixmix:
Node 18 failed - didn't download the data needed. Re-running to see if it passes

